PR #84, opened 2019-09-05
Burst Redis Cache (specific post and entire cache)

renatonascalves:


services/cacheService.js
```diff
@@ -61,4 +62,4 @@ const getService = () => {
   return defaultService;
 };
 
-export default getService;
+module.exports = getService;
```
ostowe: In line with my comments above, have this function return the `defaultService` if redis isn't configured, return the results of `getRedisService` if it is.

services/cacheService.js
```diff
@@ -1,15 +1,19 @@
...
+  del: () => null,
 };
+let service;
 
 /**
  * @typedef {object} CacheService
  * @property {function} get
  * @property {function} set
+ * @property {function} del
+ * @property {function} wipe
```
renatonascalves:
```suggestion
```

PR #85, opened 2019-09-09
Allow overrides of...

ostowe:
- `<App />`
- `<ErrorMessage />`
- `app.ejs`
- `error.ejs`

Refs https://alleyinteractive.atlassian.net/browse/IRV-145 and https://alleyinteractive.atlassian.net/browse/IRV-146


package.json
```diff
@@ -1,64 +1,25 @@
 {
   "name": "irving",
-  "version": "1.0.0",
+  "private": true,
```
ostowe: Necessary to prevent issues with publishing root package (which should never be published anyway)

package.json
```diff
@@ -1,64 +1,25 @@
...
-  "homepage": "https://github.com/alleyinteractive/irving#readme",
-  "engines": {
-    "node": "10",
-    "npm": "6"
+  "author": "Alley (Alley.co)",
+  "main": "main.js",
+  "scripts": {
+    "build": "BABEL_ENV=app lerna run build",
+    "preinstall": "npx check-node-version --package",
+    "install": "lerna bootstrap --hoist --nohoist=@irvingjs/*",
```
ostowe: this will prevent hoisting of internal packages (which I think is an lerna bug...)

packages/core/config/irving.config.default.js
```diff
@@ -6,7 +6,7 @@ import Placeholder from 'components/placeholder';
 import withLoader from 'components/hoc/withLoader';
 
 const irvingDefaultConfig = {
-  packages: {},
+  packages: [],
```
ostowe: Packages don't need to have keys (but they may need to include a package name, in the future)

packages/core/cli/start.js
```diff
@@ -14,19 +14,19 @@ const debug = createDebug('server:error');
...
+const { maybeRequireUserModule } = require('../utils/userModule');
 
 const {
   PORT = 3001,
   NODE_ENV,
 } = process.env;
 const app = express();
 
-app.set('views', path.join(__dirname, '../server/views'));
+// app.set('views', path.join(__dirname, '../server/views'));
```
zrothauser: Remove?

PR #119, opened 2019-12-21
Add action creator to reset the loaded default component data

zrothauser:


zrothauser:
Closing in favor of #120

PR #122, opened 2019-12-29
Url Updates

ostowe:
This PR will:

* Fix issue with `getRelativeUrl` function erroneously removing protocols like `mailto:`
* Add support for creating passthrough proxies directly in your `irving.config.server.js` file

### DIFF
--- a/packages/core/utils/getRelativeUrl.js
+++ b/packages/core/utils/getRelativeUrl.js
@@ -35,6 +35,7 @@ export default function getRelativeUrl(url) {
     /**
      * Consider provided target URL to be relative (and transform/return it accordingly) only if:
      * - Current hostname is included in the provided target URL
+     * - URL does not begin with a hash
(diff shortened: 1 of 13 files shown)


packages/core/utils/getRelativeUrl.js
```diff
@@ -20,15 +20,23 @@ export default function getRelativeUrl(url) {
...
-      host.includes(window.location.host) ||
-      window.location.host.includes(host)
+      (
+        host === window.location.host ||
+        host.includes(window.location.host) ||
+        window.location.host.includes(host)
+      ) &&
+      (
+        ('http:' === protocol || 'https:' === protocol) &&
+        'rss' !== url.split('.').pop()
```
jameswburke: Instead of checking for `.rss` extensions, could we run it through the regex lists defined for the proxy? In theory that would capture things more accurately.

Additionally, maybe we just check if there's a file extension at all?

PR #127, opened 2020-01-10
Cookie updates (and other things)

ostowe:
This PR will:
- Replace `react-cookie` with underlying `universal-cookie` library. We're providing whitelisted cookies to redux anyway, why set up a separate provider?
- Get cookies updates and provide to redux on client side too
- Turn cookie extraction/whitelist process into a selector to make it more reusable
- Add `process.exitCode = 1` to webpack build, causes travis to actually fail when build fails.
- Switch from `maybeResolveUserModule` to `maybeResolveBuildModule` for config files, otherwise build breaks on VIP
- Minor fix to GTM: don't use object spread for `initialRender` event, that code won't be transpiled.
- Use `gridColumns` instead of `columns` when constructing responsive styles for grid container component.


packages/core/selectors/getRouteKey.test.js
```diff
@@ -0,0 +1,58 @@
+import getRouteKey from './getRouteKey';
+
+beforeEach(() => {
+  // Allow myFunCookie to be included in routes.
+  process.env.COOKIE_MAP_LIST = 'myFunCookie';
```
pattiereaves: 🤡

PR #134, opened 2020-01-30
Add logic to configure blacklist for trailing slash logic

ostowe:
[**IRV-256**](https://alleyinteractive.atlassian.net/browse/IRV-256)


packages/core/config/irvingConfigSchema.js
```diff
@@ -22,4 +22,5 @@ module.exports = {
   stylelintConfig: [() => {}],
   styleguideConfig: [() => {}],
   styleguideSetup: [],
+  trailingSlashBlacklist: [],
```
zrothauser: See https://alleyinteractive.slack.com/archives/C8LFQ1S13/p1580421913350800 about the naming here

PR #141, opened 2020-02-18
Match the redis key to the API request

zrothauser:
https://alleyinteractive.atlassian.net/browse/IRV-250


zrothauser:
> 🗡
> 
> Do you think this change will impact in any way sites that decide to upgrade? Specifically this change:
> 
> ```
> before: response = await fetchComponents(...args);
> after: response = await fetchComponents(path, search, cookie, context);
> ```

It shouldn't, this function was really a wrapper around `fetchComponents`, which expected the arguments in that order:
https://github.com/alleyinteractive/irving/blob/master/packages/core/services/fetchComponents.js#L131

ostowe:
💯 looks good

PR #146, opened 2020-03-03
Don't convert subdomain urls to relative

ostowe:


packages/core/utils/getRelativeUrl.js
```diff
@@ -62,6 +63,7 @@ export default function getRelativeUrl(url) {
     // In case an error occurs, assume URL is not relative.
     result = false;
   }
+  // console.log(url, result);
```
amyevans:
```suggestion
```

PR #152, opened 2020-03-06
add test suite for component data reducer

cahdeemer:
[https://alleyinteractive.atlassian.net/browse/PDACG-1536](https://alleyinteractive.atlassian.net/browse/PDACG-1536)

Adds tests to the component data reducer in Irving as a part of my PD


packages/core/reducers/componentDataReducer.test.js
```diff
@@ -0,0 +1,167 @@
+import componentDataReducer from './componentDataReducer';
+import { merge } from 'lodash/fp';
```
kevinfodness: This import isn't used in this test, so you can remove it.

packages/core/reducers/componentDataReducer.test.js
```diff
@@ -0,0 +1,167 @@
...
+import {
+  REQUEST_COMPONENT_DATA,
+  RECEIVE_COMPONENT_DATA,
+  RECEIVE_COMPONENT_DATA_ERROR,
+} from 'actions/types';
+import {
+  defaultState,
+  componentDataMeta,
+} from './defaultState';
```
kevinfodness: These imports aren't used either, so you can remove them.

packages/core/reducers/componentDataReducer.test.js
```diff
@@ -0,0 +1,167 @@
...
+      },
+    });
+  });
+
+  it('Should add the data to state when using RECEIVE_COMPONENT_DATA and data exists', () => {
+    const state = {
+    testEndpoint: {
+        error: false,
+        loaded: true,
+        loading: false,
```
kevinfodness: Does `RECEIVE_COMPONENT_DATA` change the value of `loaded` and `loading`? If so, you should update them here so that `loaded` is `false` and `loading` is `true`.

packages/core/reducers/componentDataReducer.test.js
```diff
@@ -0,0 +1,167 @@
...
+      },
+    });
+  });
+
+  it('Should add error to state when there is an error in the payload', () => {
+    const state = {
+      testEndpoint: {
+        error: false,
+        loaded: true,
```
kevinfodness: Same as above.

packages/core/reducers/componentDataReducer.test.js
```diff
@@ -0,0 +1,167 @@
...
+        loading: false,
+      },
+    };
+    const error = new Error('foo');
```
kevinfodness: Nice

PR #165, opened 2020-04-11
getRequestUrl Improvements

renatonascalves:
Refactoring and adding tests for the `getRequestUrl` function.


packages/core/components/hoc/withData/getRequestUrl.js
```diff
@@ -1,26 +1,17 @@
-import URL from 'url-parse';
-
 const getRequestUrl = (endpoint) => {
-  let requestUrl;
+  switch (true) {
```
davisshaver, line 2: 💡

PR #170, opened 2020-04-11
Adding describe to more functions/helpers

renatonascalves:


some-feature.diff
```diff
@@ -0,0 +1,530 @@
+diff --git a/packages/core/components/helpers/button/index.js b/packages/core/components/helpers/button/index.js
```
amyevans, line 1: This file shouldn't be committed 😄

PR #174, opened 2020-05-01
User-themable styled components

ostowe:
## Relates to or closes...
[**IRV-266**](https://alleyinteractive.atlassian.net/browse/IRV-266)

* Add a new `createWithUserTheme` higher-order component
  * This HOC will pre-wrap a core component in `withThemes`, setting the component's core styles as the default theme, and return a function.
  * The user can use the returned function to pass in their own `themeMap` object to add themes to the core component (something like `const Image = themeImage({ myTheme: themeStyles });`)
  * Optionally, the user can override the default theme and core styles.
* Adds a test for this new HOC
* Fix issue with testing code that imports a module from another irving package (and that module does not get transpiled by `babel-jest`)

* I've installed and tested it on a client project (`v2.6.1-alpha.1`)
* I've written a unit test for the new HOC's functionality


packages/styled/components/image/index.js
```diff
@@ -143,15 +150,22 @@ Image.propTypes = {
       media: PropTypes.string.isRequired,
     })
   ),
+  /**
+   * Theme for this component
```
renatonascalves, line 154:
```suggestion
   * Theme for this component.
```

PR #181, opened 2020-05-12
Updates to Redis service and cache purging logic

ostowe:
[**IRV-262**](https://alleyinteractive.atlassian.net/browse/IRV-262)

* Update the `cacheService` to be more forgiving:
  * Allow usage of either `REDIS_MASTER` or `REDIS_URL` for non-vip environments.
  * Draw down some code from VIP's Redis implementation including options for a `REDIS_PASSWORD` and retry strategy.
  * Allow a `redis://` protocol in redis host
* Rename purge functions
* Refactor `purgeEndpointCache` to receive any path, convert it into a redis key. This allows us to use the same `PURGE` request for both VIP Go and irving core.
* Add more tests to redis service and purge functions
* Update all tests to use `__mocks__` directories.

I've tested it on the Irving Develop environment and, soon, a client project as well.


packages/core/utils/shouldAuthorize.js
```diff
@@ -0,0 +1,23 @@
+import get from 'lodash/fp/get';
+import isNode from './isNode';
+
+/**
+ * Retrieve the bearer token cookie.
+ *
+ * @param {object} cookie Object containing valid cookies for the app.
+ */
+export const getBearerToken = (cookie) => (
```
kevinfodness, line 9: This seems like something that could either be included directly in the function below (e.g., `const authorizationBearerToken = get('authorizationBearerToken', cookie)` or split out into its own file as a default export.

packages/core/services/utils/getRedisOptions.test.js
```diff
@@ -0,0 +1,39 @@
...
+
+// Set up the mocked cache service.
+describe('cacheService', () => {
+  beforeEach(() => {
+    process.env.REDIS_URL = '';
+    process.env.REDIS_MASTER = '';
+  });
+
+  it('should allow usage of REDIS_URL env var', () => {
+    process.env.REDIS_URL = '127.0.0.1:6379';
```
amyevans, line 11: Fwiw this URL doesn't work for me locally, `192.168.50.4:6379` is what works. Since it's just a test I know it doesn't really matter, but it _could_ potentially be confusing if our devs are looking to tests for possible real-world values.

kevinfodness: The `192.168.50.4` address is what we use for Broadway and is Alley-specific. I *think* that if this is running in production, Redis would be located at the local loopback address. Alternately, I've used a dummy address of `123.456.789.0` before, which is both a real IP and also obviously test data.

PR #182, opened 2020-05-12
Support to the express-naked-redirect package

renatonascalves:
One can add support to the https://www.npmjs.com/package/express-naked-redirect package via the Irving config.

This is how one would add support in `irving.config.server.js`:

```
const styledComponentsConfig = require('@irvingjs/styled');

const config = {
  name: 'irving-dev-app',
  packages: [
    styledComponentsConfig,
  ],
  customizeRedirect: {
    subDomain: 'cool',
    protocol: 'https',
  },
};

module.exports = config;
```

Imagine we have a site named `http://interesting.site`. With the above config, that would be redirected to `https://cool.interesting.site`.

---

* There is no default value or config.
* The redirect will only be applied if supported fields are added via the `nakedRedirect` config key.
* The full list of fields supported can be seen [here](https://www.npmjs.com/package/express-naked-redirect).
* We are not adding any custom setup/redirect.
* The related packages are only loaded if the config field is available.

---

It is important to understand that to use this functionality, a few things need to be set up **beforehand**. In the example provided, if the `cool` subdomain was not set up, the redirect will happen but the page will load with an error since it won't exist.

The same is true for `http`, `https` or `www`. The integration only takes care of redirecting but the support, or lack of, should be a prior task before activating this.


packages/core/server/nakedRedirect.js
```diff
@@ -0,0 +1,28 @@
+const getConfigField = require('../utils/getConfigField');
+
+const nakedRedirect = (req, res, next) => {
+  const nakedRedirectConfig = getConfigField('nakedRedirect');
+
+  if (nakedRedirectConfig) {
+    let expressNakedRedirect;
+    try {
+      // Require the package.
+      expressNakedRedirect = require('express-naked-redirect'); // eslint-disable-line global-require
```
kevinfodness, line 10: Can you do the global require above, but just not call the function if the conditions aren't met? If that approach doesn't produce side effects, I'd recommend doing it that way.

packages/core/server/nakedRedirect.js
```diff
@@ -0,0 +1,28 @@
...
+
+      // Only load this package if used.
+      if (nakedRedirectConfig.except) {
+        require('url-pattern'); // eslint-disable-line global-require
```
kevinfodness, line 14: Same here - does including this package produce side effects if it's not used?

renatonascalves:
@kevinfodness Could you take another look? Since I tried a simpler way.

PR #184, opened 2020-05-14
Fix logging

ostowe:
N/A

* Import environmental variables earlier in `start.js` to get the `debug` package to work again
* Fix both core and vip log services to use newrelic properly

Tested in irving dev environment.


packages/core/cli/start.js
```diff
@@ -1,29 +1,32 @@
...
-const express = require('express');
-const proxy = require('http-proxy-middleware');
-const cookiesMiddleware = require('universal-cookie-express');
+// Get environmental variables
+const getEnv = require('../config/env');
+const {
+  API_ROOT_URL,
+  API_ORIGN,
+  NODE_ENV,
+} = getEnv();
```
xmattus, line 8: Will these be ok without defaults?

packages/core/services/fetchComponents.js
```diff
@@ -120,7 +120,11 @@ async function cachedFetchComponents(
     context
   );
   const key = `components-endpoint:${componentsQuery}`;
-  const info = { cached: false, route: key };
+  const info = {
+    cached: false,
+    endpoint: `${process.env.API_ROOT_URL}/components?${componentsQuery}`,
```
xmattus, line 125: Same as above. Presumably `API_ROOT_URL` will almost always be set though.

packages/core/cli/start.js
ostowe: API_ORIGIN will fall back to API_ROOT_URL, and yeah I figure the app _should_ break if that env var isn't set, not sure what it would even look like if it wasn't.

PR #185, opened 2020-05-15
AWS Lambda Package

renatonascalves:
Docs: https://github.com/alleyinteractive/irving/wiki/Deploying-on-AWS-Lambda


packages/core/server/index.js
```diff
@@ -38,7 +37,7 @@ const passthrough = createProxyMiddleware({
   changeOrigin: true,
   followRedirects: true,
   secure: 'development' !== process.env.NODE_ENV,
-  target: API_ORIGIN || API_ROOT_URL.replace('/wp-json/irving/v1', ''),
+  target: process.env.API_ORIGIN || process.env.API_ROOT_URL.replace('/wp-json/irving/v1', ''),
```
davisshaver, line 40: any benefits to removing the destructuring here?

renatonascalves: Honestly, not sure. But it was complaining the variable didn't exist when I was testing locally. Like it couldn't find it.

davisshaver: Gotcha - not a blocker! Destructure defaults might be able to fix the eslint warning though

PR #196, opened 2020-05-27
Move server creation into a new server/index.js file

ostowe:
[**IRV-327**](https://alleyinteractive.atlassian.net/browse/IRV-327)

- Move express server creation to its own file so it can be imported independent of the server start process.
- Add a new `preExportServer` config field to modify app just before it's exported.
- Add new `basicAuth` config field that must be `true` for basic auth to be applied to the production app, in addition to the username/password env vars

- Tested in irving dev environment.


packages/core/server/index.js
```diff
@@ -0,0 +1,78 @@
...
+
+// Create proxies for each configured proxy pattern.
+proxyPassthrough.forEach((pattern) => {
+  app.use(pattern, passthrough);
+});
+
+// Add universal cookies middleware.
+app.use(cookiesMiddleware());
+
+// Naked Redirect.
```
renatonascalves, line 54:
```suggestion
// Customize Redirect.
```

packages/core/server/index.js
```diff
@@ -0,0 +1,78 @@
...
+app.use(customizeRedirect());
+
+if ('development' === process.env.NODE_ENV) {
```
renatonascalves, line 57:
```suggestion
if ('development' === NODE_ENV) {
```

packages/core/server/index.js
```diff
@@ -0,0 +1,78 @@
+/* eslint-disable global-require, no-console, import/order */
+const {
+  API_ROOT_URL,
+  API_ORIGIN,
+} = process.env;
```
renatonascalves, line 5:
```suggestion
const {
  API_ROOT_URL,
  API_ORIGIN,
  NODE_ENV,
} = process.env;
```

packages/core/server/index.js
```diff
@@ -0,0 +1,78 @@
...
+// Run all customize server functions.
+const irvingServerMiddleware = getConfigField('customizeServer');
+irvingServerMiddleware.forEach((middleware) => middleware(app));
+
+// Set up a reusable proxy for responses that should be served directly.
+const proxyPassthrough = getConfigArray('proxyPassthrough');
+const passthrough = createProxyMiddleware({
+  changeOrigin: true,
+  followRedirects: true,
+  secure: 'development' !== process.env.NODE_ENV,
```
renatonascalves, line 40:
```suggestion
  secure: 'development' !== NODE_ENV,
```

packages/core/server/index.js
```diff
@@ -0,0 +1,78 @@
...
+  log.error(err);
+
+  if (res.headersSent) {
+    return next(err);
+  }
+
+  return res.sendStatus(500);
+});
+
+// Run all export server functions.
```
renatonascalves, line 74:
```suggestion
// Run all pre-export server functions.
```

packages/core/server/index.js
```diff
@@ -0,0 +1,78 @@
...
+const irvingServerMiddleware = getConfigField('customizeServer');
+irvingServerMiddleware.forEach((middleware) => middleware(app));
+
+// Set up a reusable proxy for responses that should be served directly.
+const proxyPassthrough = getConfigArray('proxyPassthrough');
+const passthrough = createProxyMiddleware({
+  changeOrigin: true,
+  followRedirects: true,
+  secure: 'development' !== process.env.NODE_ENV,
+  // @todo make this not specific to WP eventually.
```
renatonascalves, line 41: Doesn't the use of `API_ORIGIN` takes care of the WP specificity?

packages/core/server/index.js
ostowe: Unfortunately (and perhaps interestingly) this isn't something we can do in this case. Webpack will look for exactly `process.env.NODE_ENV` when performing static replacement of that variable with its value and closing of execution contexts. Take a look here: https://webpack.js.org/plugins/environment-plugin/

packages/core/server/index.js
ostowe: I guess so yeah, so maybe I can remove this todo.

packages/core/server/index.js
renatonascalves: Interesting!

PR #202, opened 2020-06-02
Styled components package

xmattus:
This is the feature branch for the Irving core library of components. The components are currently located in `candidate-components` but will eventually be merged with or replace entirely the `components` directory.

This is a companion to https://github.com/alleyinteractive/wp-irving/pull/169

--------------------

* Moves the first set of React components from the candidates into this standalone package.


packages/styled-components/components/menu/themes/default.js
```diff
@@ -0,0 +1,44 @@
...
+  justify-content: center;
+  width: 100%;
+`;
+
+// Wrapper around the menu name.
+export const NameWrapper = styled.h3`
+  display: none;
+`;
+
+export const Inner = styled.ul`
```
pattiereaves, line 16: Considering that a menu is "ordered," an `ol` may make more sense semantically.

packages/styled-components/components/menu/themes/default.js
```diff
@@ -0,0 +1,44 @@
...
+  align-items: center;
+  display: flex;
+  justify-content: center;
+  list-style: none;
+`;
+
+export const ItemWrapper = styled.span`
```
pattiereaves, line 23: [The list item element must be a direct child](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ul) of a UL/OL, this needs to be refactored.

packages/styled-components/components/menu/menuItem.js
```diff
@@ -0,0 +1,100 @@
...
+
+  const {
+    ItemWrapper,
+    Dropdown,
+  } = theme;
+
+  return (
+    <ItemWrapper key={id} classnames={classes}>
+      <li>
+        <a href={url} target={target} title={attributeTitle}>{title}</a>
```
pattiereaves, line 33: Title attributes aren't supported: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a
```suggestion
        <a href={url} target={target}>{title}</a>
```

packages/styled-components/components/menu/themes/default.js
```diff
@@ -0,0 +1,44 @@
...
+      text-decoration: none;
+    }
+
+    a:hover {
+      text-decoration: underline;
+    }
+  }
+`;
+
+export const Dropdown = styled.ul`
```
pattiereaves, line 41: Ditto the `ol` comment here.

packages/styled-components/components/menu/themes/defaultVertical.js
```diff
@@ -0,0 +1,48 @@
...
+
+/* eslint-disable import/prefer-default-export */
+export const Wrapper = styled.div`
+  display: flex;
+  flex-direction: column;
+  width: 100%;
+`;
+
+// Wrapper around the menu name.
+export const NameWrapper = styled.h3`
```
pattiereaves, line 11: Depending on where a menu is used the required heading level can change -- keeping it out of the style sheet as a dependency is a good idea.

packages/styled-components/components/menu/themes/defaultVertical.js
```diff
@@ -0,0 +1,48 @@
...
+
+export const Inner = styled.ul`
+  display: flex;
+  flex-direction: column;
+  margin: 0;
+  padding: 0;
+  list-style: none;
+`;
+
+export const ItemWrapper = styled.span`
```
pattiereaves, line 25: Ditto the li/ol comments in this theme.

packages/styled-components/components/link/themes/default.js
```diff
@@ -0,0 +1,13 @@
...
+
+/* eslint-disable import/prefer-default-export */
+export const LinkWrapper = styled.a`
+  color: black;
+  font-family: inherit;
+  text-decoration: none;
+
+  &:hover {
+    text-decoration: underline;
+  }
```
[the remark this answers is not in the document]
jameswburke: Good call. I don't have a solid answer for what it should look like yet, so I added a @todo and will let a larger group have that discussion soon.

packages/styled-components/components/logo/themes/default.js
```diff
@@ -0,0 +1,20 @@
+import styled from 'styled-components';
+
+/* eslint-disable import/prefer-default-export */
+export const LogoWrapper = styled.span`
+  display: block;
+
+  img {
+    max-width: 250px;
```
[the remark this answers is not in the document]
jameswburke: 99% of this will be overridden by default anyway, so I just picked a value. Added a @todo for future discussion.

packages/styled-components/components/html/themes/default.js
```diff
@@ -0,0 +1,390 @@
...
+  }
+
+  a button,
+  button,
+  input[type="submit"],
+  input[type="reset"],
+  input[type="button"] {
+    font-size: 1rem;
+    display: inline-block;
+    padding: 6px 12px;
```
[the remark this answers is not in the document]
jameswburke: Added a todo. I plucked this from some start HTML thing. It needs to be reviewed by a group top to bottom to make decisions about.

packages/styled-components/components/menu/themes/defaultVertical.js
jameswburke: This can be overridden by a developer injecting a new theme, which I believe would be the recommended solution.

packages/styled-components/components/menu/menuItem.js
jameswburke: Thanks. Moved it to the <li> element.

PR #210, opened 2020-06-13
Updates to text component, and adds query components.

jameswburke:
IRV-410 and IRV-412

* Refactor `irving/html` into `irving/text` to clean up responsibilities and make development easier.
* Update `irving/text` to support plaintext, html, and oembed.
* Basic themes for `irving/text` (h1-h6 & caption).
* Adds `irving/pagination` component to render pagination UI.
* Adds `irving/searchForm` component to render a search form.

Tested in the example theme. Still needs unit tests.


packages/styled-components/components/pagination/index.js
```diff
@@ -0,0 +1,223 @@
...
+
+/**
+ * Using the same props from a Pagination component, build the url for a given
+ * page.
+ *
+ * @param {object}  props Pagination component props.
+ * @param {integer} page  Page for the url.
+ * @return {string}
+ */
+const buildUrl = (props, page) => {
```
jameswburke, line 17: This setup isn't perfect, but I'm comfortable with it as an alpha component.

Suggestions welcome, but they may get bumped to the next review stage.

packages/styled-components/components/pagination/index.js
```diff
@@ -0,0 +1,223 @@
...
+ */
+const Pagination = (props) => {
+  const {
+    // eslint-disable-next-line no-unused-vars
+    baseUrl,
+    currentPage,
+    displayFirstAndLast,
+    displayPrevAndNext,
+    // eslint-disable-next-line no-unused-vars
+    paginationFormat,
```
jameswburke, line 49: Doing this and the other `disable` so that we can still validate the type + get a default, but they're only used in the context of `buildUrl()`.

packages/styled-components/components/text/index.js
```diff
@@ -1,53 +1,113 @@
...
   default: defaultStyles,
+  html: htmlStyles,
+  unstyled: unstyledStyles,
+  caption: captionStyles,
+  h1: h1Styles,
+  h2: h2Styles,
+  h3: h3Styles,
+  h4: h4Styles,
+  h5: h5Styles,
+  h6: h6Styles,
```
jameswburke, line 110: None of these are final. Just a simple first-pass before some frontend devs can make some better decisions around it.

packages/styled-components/components/pagination/index.js
jameswburke: Agreed on both fronts. Adding a todo.

packages/styled-components/components/searchForm/index.js
```diff
@@ -0,0 +1,97 @@
...
+        ref={register}
+        type="text"
+      />
+      <SearchFormSubmitButton type="submit">Search</SearchFormSubmitButton>
+    </SearchFormWrapper>
+  );
+};
+
+SearchForm.defaultProps = {
+  baseUrl: '/',
```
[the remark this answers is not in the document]
jameswburke: `/` is actually the default in WP. Adding/using `/search/` is largely a James-pattern.

packages/styled-components/components/text/index.js
```diff
@@ -1,53 +1,113 @@
...
 const Fragment = (props) => {
   const {
-    children,
+    content,
+    html,
+    oembed,
+    style,
     tag,
-    theme,
+    theme = defaultStyles,
```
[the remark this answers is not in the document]
jameswburke: I had some issues with the Menu component where I was getting a FOUC issue w/o it, so I kept up with the pattern. We may be able to remove.

PR #214, opened 2020-06-24
Feature/tweak various components

jameswburke:
Nothing in particular.

Modify some of the Styled Components based on various conversations or issues.

* Fix style proptype.
* Re-add icons to social sharing until a better solution is found.
* Use a non-local version of the CSS reset file in `<App />`.
* Allow container to use a string value like `max_width="800px"`.

Tested locally.

### DIFF
--- a/packages/styled-components/components/container/index.js
+++ b/packages/styled-components/components/container/index.js
@@ -79,7 +79,10 @@ Container.propTypes = {
   /**
    * CSS styles.
    */
-  style: PropTypes.object,
+  style: PropTypes.oneOfType([
+    PropTypes.array,
+    PropTypes.object,
+  ]),
   /**
    * Tag used to render.
    */
--- a/packages/styled-components/components/container/index.js
+++ b/packages/styled-components/components/container/index.js
@@ -17,6 +17,8 @@ const widths = {
  * @todo Is there a better way to handle the maxWidth and background image
  *       functionality? We want this functionality to be independent from
  *       themes.
+ * @todo If children is empty, should we render anything? Do we want a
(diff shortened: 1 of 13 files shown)


packages/styled-components/components/container/index.js
```diff
@@ -17,6 +17,8 @@ const widths = {
  * @todo Is there a better way to handle the maxWidth and background image
  *       functionality? We want this functionality to be independent from
  *       themes.
+ * @todo If children is empty, should we render anything? Do we want a
```
tylerreckart, line 20: I would lean towards not rendering anything at all in this case. IMO it doesn't make sense to send empty markup to the DOM that won't be used in any meaningful way.

packages/styled-components/components/container/index.js
```diff
@@ -31,10 +33,12 @@ const Container = (props) => {
   const { ContainerWrapper } = theme;
 
   // Setup a max-width style.
-  if ('string' === typeof maxWidth) {
-    style.maxWidth = `${widths[maxWidth]}px`;
+  if ('string' === typeof maxWidth && undefined !== widths[maxWidth]) {
+    style.maxWidth = `${widths[maxWidth]}px`; // Use one of our predefined sizes.
+  }else if ('string' === typeof maxWidth ) {
```
tylerreckart, line 38:
```suggestion
  } else if ('string' === typeof maxWidth ) {
```

packages/styled-components/components/container/index.js
jameswburke: Cool. I agree, implemented.

PR #216, opened 2020-06-24
feat(image component.): implement the first alpha of an Image component

jameswburke:
Jira's [IRV-475](https://alleyinteractive.atlassian.net/browse/IRV-475).

Implement the first Image alpha component, derived from the `@irvingjs/styled/components/image`.

Tested locally and documented in Storybook.

### DIFF
--- a/packages/styled-components/components/image/themes/default.js
+++ b/packages/styled-components/components/image/themes/default.js
@@ -23,13 +23,11 @@ const getAspectRatioStyles = (props) => {
   return '';
 };
 
-/* eslint-disable import/prefer-default-export */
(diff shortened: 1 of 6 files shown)


packages/styled-components/components/image/themes/default.js
```diff
@@ -0,0 +1,69 @@
...
+      position: absolute;
+      top: 0;
+      width: 100%;
+    `;
+  }
+
+  return '';
+};
+
+/* eslint-disable import/prefer-default-export */
```
tylerreckart, line 26: IIRC you shouldn't need to disable the default export rule if you're exporting multiple variables in a single file.

PR #220, opened 2020-06-26
Feature/site theme

jameswburke:
[IRV-486](https://alleyinteractive.atlassian.net/browse/IRV-486)

 This commit implements a new SiteThemeProvider which acts as a middleware component between the siteTheme provider setup in the backend, and allows those theme values to be set in css values set by the `style` prop.

Tested locally.

### DIFF
--- a/packages/styled/components/hoc/withThemes.js
+++ b/packages/styled/components/hoc/withThemes.js
@@ -1,18 +1,34 @@
-import React from 'react';
+import React, { useContext } from 'react';
 import PropTypes from 'prop-types';
 import getDisplayName from '@irvingjs/core/utils/getDisplayName';
 import assign from 'lodash/fp/assign';
+import get from 'lodash/get';
+import { ThemeContext } from 'styled-components';
 
 const withThemes = (themeMap) => (WrappedComponent) => {
   const ThemedComponent = (props) => {
-    const { themeName } = props;
+    const { themeName, style } = props;
     const theme = assign(
       themeMap.default,
       themeMap[themeName]
     );
 
+    // If the value of a property in a styled components `style` property is
+    // available in the styled component theme, use that value. This allows us
+    // to do things like `color: theme.colors.black` and have it become
+    // `color: #000000`.
+    const themeContext = useContext(ThemeContext);
+    const newProps = props;
+    Object.keys(style).forEach((property) => {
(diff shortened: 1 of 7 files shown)


packages/styled/components/hoc/withThemes.js
```diff
@@ -1,18 +1,34 @@
...
+    // `color: #000000`.
+    const themeContext = useContext(ThemeContext);
+    const newProps = props;
+    Object.keys(style).forEach((property) => {
+      newProps.style[property] = get(
+        themeContext,
+        newProps.style[property],
+        newProps.style[property]
+      );
+    });
```
pattiereaves, line 28: A reducer makes it more clear what the themeContext will replace in the style prop.

```suggestion
const newStyle = Object.keys(style)
  .reduce((acc, property) => {
    const themeStyles = get(
      themeContext,
      style[property],
      style[property] // is this right, should the key be the default value?
    );
    
    return {
      ...acc,
      [property]: themeStyles,
    };
  }, style);
const newProps = { ...props, style: newStyle };
```

packages/styled/components/hoc/withThemes.js
```diff
@@ -21,10 +37,18 @@ const withThemes = (themeMap) => (WrappedComponent) => {
      * Prop indicating which theme to use.
      */
     themeName: PropTypes.oneOf(Object.keys(themeMap)),
+    /**
+     * CSS style.
+     */
+    style: PropTypes.oneOf([
+      PropTypes.array,
```
pattiereaves, line 44: Will this still work if style is an array?

jameswburke: It's an issue on the PHP side,

An empty (`[]`) or indexed (`[0=>'', 1=>'']`) array will be an array in JSON, and associate arrays (`['hello'=>'world', 'foo'=>'bar']`) will be an object in JSON. So if `style` is empty or unset on the WP side, it comes over as an array, which throws a propType error.

We're fixing on the WP side, but it's part of a much bigger thing and isn't ready yet. So you're right to call it out, but it's a temporary thing.

packages/styled/components/hoc/withThemes.js
jameswburke: Thanks, this makes sense.

PR #226, opened 2020-07-02
feat(text component - responsive video theme): adds a new responsiveVideo theme to irving/text

jameswburke:
If the HTML provided by an instance of the irving/text component contains a video (such as an iframe
embed), then the `responsiveVideo` theme will render it responsively, with an optional aspect ratio.

None.

Add support for a basic responsive video embed option.

Tested locally.


packages/styled-components/components/text/index.js
```diff
@@ -6,6 +6,7 @@ import { richText } from '@irvingjs/core/config/html';
 import withThemes from '@irvingjs/styled/components/withThemes';
 import * as defaultStyles from './themes/default';
 import * as htmlStyles from './themes/html';
+import * as responsiveVideoStyles from './themes/responsiveVideo';
```
amyevans, line 9: Since it doesn't _have_ to be a video, worth considering maybe "responsive embed" as the name?

PR #233, opened 2020-07-11
Add examples directory with an example package

ostowe:
N/A

Add an examples directory with a sample package you could use as a starter if you want to create your own.


examples/package/package.json
```diff
@@ -0,0 +1,34 @@
...
+  ],
+  "main": "index.js",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/alleyinteractive/irving/packages/example-package"
+  },
+  "author": "Alley (Alley.co)",
+  "license": "MIT",
+  "scripts": {
+    "test": "jest --passWithNoTests",
```
cahdeemer, line 17: TIL about the passwithNoTests flag in Jest and which Irving packages use it :-)

PR #236, opened 2020-07-16
Forms update

ostowe:
N/A

* Deprecate our usage of the `react-use-form-state` hook and replace it with `react-hook-form`
* Remove all the components in this package, they were kinda superfluous and difficult to maintain before and are now extra superfluous with `react-hook-form`
* Focus solely on connected forms (forms submitting to an external endpoint). Non-connected forms can easily be set up solely using `react-hook-form`.
* Convert connected forms logic to a hook.

I've tested in a client project


ostowe:
Agreed. I'm in the process of updating the documentation at least, which is a start.

PR #245, opened 2020-07-21
Config Overhaul

ostowe:
* Remove `irving.config.server.js`
* Transition the following config fields (listed by their expected path) to be discovered via filesystem while app is bootstrapping (instead of using `irving.config.server.js`)
  * `server/customizeServer`
  * `server/customizeProdServer`
  * `server/customizeDevServer`
  * `server/exportServer`
  * `server/startServer`
  * `server/customizeRedirect`
  * `config/envAllowlist`
  * `config/proxyPassthrough`
  * `config/webpack.config.js`
  * `config/postcss.config.js`
  * `config/babel.config.js`
  * `services/logService`
  * `services/cacheService`
  * `services/cacheClient`
  * `services/monitorService`
* Add two new config fields:
  * `config/webpack.config.client.js` - if you only want to configure the client webpack config.
  * `config/webpack.config.server.js` - if you only want to configure the server webpack config.
* Unify process of merging config values
  * If config default is an array, configured values will be spread into the array after the default values.
  * If config default is an object, configured values will be spread into the object after default values.
  * If config default is an array with its first values set to a function, configured functions will be added to the array without being called.
  * If config default is a function, the most recently discovered config function will be used (with the user's function taking precedence)
  * If config default is an object or array and configured value(s) are functions, those functions will be called with the current config value and are expected to return a config value of the same type.
* Services will now:
  * Be configured to more reliably use the default service
  * Default services will be kept in a separate file
  * Services that cannot be run in a browser context will automatically be left out of the client bundle
* Rename `process.env.WEBPACK_BUILD` to `process.env.IRVING_EXECUTION_CONTEXT` and contain the full context value, which will be one of: `development_server`, `development_client`, `production_server`, `production_client`
* Switch from `svg-react-loader` to `@svgr/webpack` because it's more stable and feature-rich and is a drop-in replacement.
* Update all packages to support new config setup

I've written a lot more unit tests and begun testing in a client project.


packages/postcss/config/stylelint.config.js
```diff
@@ -51,9 +51,6 @@ module.exports = {
     'selector-combinator-space-before': 'always',
     'selector-descendant-combinator-no-non-space': true,
     'selector-list-comma-newline-after': 'always',
-    'selector-max-specificity': ['0,2,1', {
```
joemcgill, line 54: Do we need to update a sane default for this? Are we running into specificity issues where we're writing too specific styles? Curious to know more about this change.

packages/vip-go/services/logService.js
```diff
@@ -24,10 +13,17 @@ const defaultService = {
  */
 const getService = (namespace) => {
   const env = process.env.NODE_ENV;
-  let service;
+  let service = defaultService;
```
ostowe, line 16: check that this uses core's logging as a fallback (and maybe write a test to that effect as well)

packages/vip-go/services/monitorService.js
```diff
@@ -12,36 +10,31 @@ const defaultService = {
...
-        start: () => {},
-        logError(err) {
-          client.noticeError(err);
-        },
-        logTransaction(method, status, category) {
-          client.setTransactionName(`${method} ${status} ${category}`);
-        },
-      };
-    } catch (err) {
+    // VIP Go's package can return nothing if newrelic is not instsalled or configured improperly.
```
ostowe, line 20: `installed`

packages/styled/config/stylelint.config.js
```diff
@@ -56,9 +60,6 @@ module.exports = {
     'selector-combinator-space-before': 'always',
     'selector-descendant-combinator-no-non-space': true,
     'selector-list-comma-newline-after': 'always',
-    'selector-max-specificity': ['0,2,1', {
```
ostowe, line 59: add this back in

PR #247, opened 2020-07-21
Prevent our DOM shim functionality from polluting the global scope

ostowe:
https://alleyinteractive.atlassian.net/browse/IRV-446
https://alleyinteractive.atlassian.net/browse/IRV-467

* Differentiate between node and web when setting up config for preset env. This will significantly reduce the size of the server bundle.
* Split babel configs into node, web, and test
* Add babel plugin for transforming references to the `window` global to reference our DOM shim on the server side
* Fix stylelint, fix a bunch of CSS issues now that it's working

Tested in irving dev.


packages/core/config/webpack/rules.js
```diff
@@ -106,29 +107,34 @@ module.exports = function getRules(context) {
...
-                filepath.includes('query-string') ||
-                filepath.includes('split-on-first') ||
-                filepath.includes('strict-uri-encode') ||
-                filepath.includes('abort-controller') ||
-                filepath.includes('event-target-shim')
-              )
-          ),
-        ],
-      },
+      // resource: {
```
ostowe, line 110: Removing this will probably break IE11. We might be able to get around this with a separate loader, but it's kind of annoying.

packages/styled-components/components/image/index.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import sanitizeHtml from 'sanitize-html';
-import useImageSize from '@use-hooks/image-size';
+// import useImageSize from '@use-hooks/image-size';
```
ostowe, line 4: This is not isomorphically safe, so we'll probably have to figure out a different way of getting width.

PR #251, opened 2020-07-30
Pico integration

jameswburke:
IRV-516

Add a [Pico](https://trypico.com/) integration component to Irving's integrations package.

Tested locally.


packages/integrations/components/pico/index.js
```diff
@@ -0,0 +1,59 @@
...
+
+  // On component hydration, add an event listener to watch for the script's init event.
+  useEffect(() => {
+    window.document.addEventListener(
+      'pico-init',
+      () => {
+        setPicoLoaded(true);
+      }
+    );
+  });
```
ostowe, line 31: Provide a `[]` as the second parameter here to avoid unnecessarily calling this hook again.

packages/integrations/components/pico/index.js
```diff
@@ -0,0 +1,59 @@
...
+    url: window.location.href,
+  };
+
+  const [picoLoaded, setPicoLoaded] = useState(false);
+
+  // On component hydration, add an event listener to watch for the script's init event.
+  useEffect(() => {
+    window.document.addEventListener(
+      'pico-init',
+      () => {
```
ostowe, line 27: Assign this handler to a variable, return a cleanup function. Example: 

```javascript
useEffect(() => {
    const handler = () => {
        setPicoLoaded(true);
    };
    window.document.addEventListener('pico-init', handler);
    
    return () => document.removeEvemntListener('pico-init', handler);
}, []);
```

PR #256, opened 2020-08-04
fix(config): fix configurations for forms, audio-player packages

ostowe:
No

* Fix issues with configuration fields for the `forms` and `audio-player` packages
* Update audio player reducer to use new default state object (instead of function)
* Use hooks for all redux state instead of `mapDispatchToProps` or `mapStateToProps`

Tested on irving dev


packages/forms/sagas/formSaga.js
```diff
@@ -11,6 +11,8 @@ export default function* watchRequestSubmit(data) {
     payload: { formEndpoint, submission },
   } = data;
 
+  console.log(formEndpoint);
+
```
jameswburke, line 15:
```suggestion
```

PR #258, opened 2020-08-04
IRV-592: Integrations Manager

tylerreckart:
**Important:** This PR should be reviewed in tandem with alleyinteractive/wp-irving#215, as it includes core functionality on the WP side that is required for this integration to work in its full capacity.

This PR is for the front-end work outlined in [IRV-592](https://alleyinteractive.atlassian.net/browse/IRV-592), which includes the introduction of the `<IntegrationsManager />` component and the ability to add and configure integrations through a project's `irving.config.js` file.

Configuring the integrations manager is a simple process. First, add the integrations manager component to the irving `componentMap`. Example:
```js javascript
import { IntegrationsManager } from '@irvingjs/integrations';
export default {
    'irving/integrations': IntegrationsManager,
};
```
Second, import the irving config from the integrations package and add the imported config to the `packages` section of your project's `irving.config.js` file. Example:
```js javascript
import integrationsConfig from '@irvingjs/integrations';
export default {
    packages: [
        integrationsConfig,
    ],
};
```
Importing the configuration is required in order to connect the Integration Manager to the global Redux state tree and add the `integrations` branch to the store. The integrations currently available to manage through the integrations manager include:
* `<Google Analytics />`
  * Key: `googleAnalytics`
  * Accepted Props: `trackingId`

Configurations can also be managed manually through your irving configuration file. In order to do so, all you have to do is specify the integration's key and properties to be passed into its configuration. Example:
```js javascript
import integrationsConfig from '@irvingjs/integrations';
export default {
    packages: [
        integrationsConfig,
    ],
    integrations: {
        googleAnalytics: {
            trackingId: 'UA-000000-0',
        },
    },
};
```
_Any configurations that are set manually in `irving.config.js` will override those that are retrieved from the REST API._

A unit test has been added under `components/integrationsManager/integrationsManager.test.js`. This test checks to ensure that once a hydrated configuration is present that the appropriate components are rendered in its child tree. There is also an additional check to ensure that if a hydrated configuration is not present, no integrations are added to the child tree.


packages/integrations/readme.md
```diff
@@ -6,3 +6,48 @@ A collection of Irving-friendly components for getting you started with common t
...
 * `<Parsely />` - Render the required JS for sending data to Parse.ly
+* `<YoastSchema />` - Render the required JS for Yoast.
+---
+### `<IntegrationsManager />`
+The integrations manager is a component that consumes a `irving/integrations` configuration retrieved from the REST API that dynamically configures, stores props for, and renders corresponding integrations based on the keys set in the configuration object.
+
+### Configuration
+Configuring the integrations manager is a simple process. First, add the integrations manager component to the irving componentMap. Example:
+```js javascript
+import { IntegrationsManager } from '@irvingjs/integrations';
```
jameswburke, line 17: Could this export be a default?

packages/integrations/components/integrationsManager/index.js
```diff
@@ -0,0 +1,144 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+// User Config.
+import userConfig from '@irvingjs/irving.config';
+// Redux.
+import { connect } from 'react-redux';
+import { actionHydrateComponents } from '../../actions';
+// Integration components.
+import GoogleAnalytics from '../googleAnalytics';
```
jameswburke, line 9: Can we abstract this part into another file? It'd be nice to manage the integrations component map from a different location.

packages/integrations/components/integrationsManager/index.js
```diff
@@ -0,0 +1,144 @@
...
+const IntegrationsManager = ({
+  integrations,
+  isHydrated,
+  hydrateComponents,
+  hydratedComponents,
+}) => {
+  // Declare a map of available integration components with a corresponding key.
+  const componentMap = [
+    { key: 'googleAnalytics', el: GoogleAnalytics },
+  ];
```
jameswburke, line 33: As noted above, I think this would make more sense in another file being imported.

packages/integrations/components/integrationsManager/index.js
```diff
@@ -0,0 +1,144 @@
...
+    const keyMap = Object.keys(integrations);
+
+    // Create an array for hydrated component keys and props.
+    const components = [];
+
+    componentMap.forEach((component, index) => {
+      // Get the component's key.
+      const { key } = component;
+
+      if (- 1 < keyMap.indexOf(key)) {
```
ostowe, line 48: could use `keyMap.includes(key)` here

packages/integrations/components/integrationsManager/index.js
```diff
@@ -0,0 +1,144 @@
...
+        // Retrieve the key/value pairs set for the integration and convert them into props.
+        // If a user config override exists, use it instead.
+        let props;
+        if ('integrations' in userConfig && key in userConfig.integrations) {
```
ostowe, line 52: Do you thin there'd be any use in allowing other irving packages to configure integrationgs? If so, you might look into using [`getConfigValues`](https://github.com/alleyinteractive/irving/blob/master/packages/core/config/irving/getValueFromConfig.js#L27) to grab the relevant values.

packages/integrations/components/integrationsManager/index.js
```diff
@@ -0,0 +1,144 @@
...
+    googleAnalytics: PropTypes.shape({
+      trackingId: PropTypes.string,
+    }),
+  }),
+  isHydrated: PropTypes.bool.isRequired,
+  hydrateComponents: PropTypes.func.isRequired,
+  hydratedComponents: PropTypes.arrayOf(PropTypes.object),
+};
+
+const mapStateToProps = (state, ownProps) => {
```
ostowe, line 116: I've been unofficially leaning toward using `react-redux` `useSelector` and `useDispatch` hooks instead of the connect functions for a couple reasons:
* A bit more concise than connect functions
* Makes it clearer what values are coming from state and what from API or user

Open to differing opinions, though, if you have some!

packages/integrations/components/integrationsManager/index.js
```diff
@@ -0,0 +1,144 @@
...
+    if (- 1 < index) {
+      const props = { key: component.key, ...component.props };
+      // Clone the element.
+      return React.createElement(componentMap[index].el, props, null);
+    }
+
+    return null;
+  };
+
+  if (0 < hydratedComponents.length) {
```
ostowe, line 89: I tend to prefer to check if `hydratedComponents.length` is a truthy value

packages/integrations/components/integrationsManager/index.js
```diff
@@ -0,0 +1,144 @@
...
+   * @param {Object} component - The component key and props.
+   */
+  const renderComponents = (component) => {
+    // Create a map of available component keys.
+    const keyMap = componentMap.map((co) => co.key);
+    const index = keyMap.indexOf(component.key);
+
+    // If the current component key exists in the key map, clone that component
+    // with the hydrated props for render.
+    if (- 1 < index) {
```
ostowe, line 80: Could use `keyMap.includes(component.key)` here too

packages/integrations/readme.md
tylerreckart: I don't think that would fit the established pattern. Right now the packages export the`config` as default and the components as individual named exports. Open to @ostowe's opinion, though.

packages/integrations/components/integrationsManager/index.js
tylerreckart: I think there would be merit there. During dev I tried using `getConfigValues` initially to no avail. IIRC it would return as undefined, so I think some further investigation would be needed there.

packages/integrations/components/integrationsManager/index.js
tylerreckart: I definitely think that the hooks are useful in a lot of applications, but IMO I think the connect functions are the better configuration in this instance. This is mainly because the selector hook can return any value, not just an object, and in this case I think we want to be pretty strict about the type safety of our props. Additionally we don't have access to `ownProps` when using the selector hook, so extra care would have to be taken there to get the same functionality. I don't think there are any detriments to the dispatch hook, but since the connect pattern is already present for `mapStateToProps`, I think it's consistent to use the `mapDispatch` function here too.

packages/integrations/components/integrationsManager/index.js
tylerreckart: True, but since we need the index value for the retrieving the component for the clone, the use here makes sense IMO

packages/integrations/components/integrationsManager/index.js
ostowe: Hmmm ok, we can investigate separately.

packages/integrations/components/integrationsManager/index.js
ostowe: Makes sense to me!

packages/integrations/components/integrationsManager/index.js
ostowe: 👍

PR #266, opened 2020-08-09
Cookie Security fix

ostowe:
[**IRV-637**](https://alleyinteractive.atlassian.net/browse/IRV-637)

* Add a new `preloadedStateDenylist` config field
* Allow users to omit certain fields from being rendered in preloaded state in server HTML
* By default prevent cookies from being rendered in server HTML

* Wrote unit tests
* Checked on Irving dev


packages/core/server/utils/encodeState.test.js
```diff
@@ -0,0 +1,22 @@
...
+      test: {
+        nested: {
+          state: 'lorem ipsum',
+        },
+      },
+    });
+    expect(JSON.parse(state).test.nested).toEqual({});
+  });
+
+  it('encode unsafe HTMl characters', () => {
```
pattiereaves, line 15:
```suggestion
  it('encodes unsafe HTMl characters', () => {
```

packages/styled-components/components/link/index.js
```diff
@@ -31,6 +32,9 @@ const Link = (props) => {
     LinkWrapper,
   } = theme;
 
+  const cookies = useSelector((state) => state.route.cookie);
+  console.log(cookies);
```
pattiereaves, line 36: Are these lines still necessary?

stories/05-architecture/01-Configuration.stories.mdx
```diff
@@ -330,6 +330,23 @@ export default {
 };
 ```
 
+## Denylist for preloaded client-side state
+* **where to configure**: `irving.config.js` in the `preloadedStateDenylist` key
+* **type**: array
+* **purpose**: Prevent certain keys in global redux state from being rendered in the server-side HTML. We follow Redux' [server-rendering guide](https://redux.js.org/recipes/server-rendering) to pass the initial state constructed in the first server-side render to the client side via a global `window.__PRELOADED_STATE__` variable. If you have both potentially sensitive data in redux state and are employing an edge cache, this presents a situation where one user's data may be presented to another user. This key will allow you to prevent this from happening and, by default, this method will be employed for passing cookie data from server to client securely.
+
+This config field expects an array of objects, in which each object has two keys: `key` and `rehydrationFunction`. The key can be in any format acceptable by lodash's [get (or set) functions](https://lodash.com/docs/4.17.15#get). The `rehydrationFunction` will be run to rehydrate data on the client side, if necessary.
```
pattiereaves, line 338:
```suggestion
This config field expects an array of objects, in which each object has two keys: `key` and `rehydrationFunction`. The key can be in any format acceptable by lodash's [get (or set) functions](https://lodash.com/docs/4.17.15#get). The `rehydrationFunction` will rehydrate data on the client side, if necessary.
```

PR #271, opened 2020-08-13
IRV-614: Add SiteStats component to integrations manager configuration

tylerreckart:
See alleyinteractive/wp-irving#220 for the backend work


packages/integrations/components/siteStats/index.js
```diff
@@ -0,0 +1,53 @@
...
+      })
+    }`;
+
+    // Set up stats gif attributes.
+    gif.src = src;
+    gif.alt = ':)';
+    gif.width = 6;
+    gif.height = 5;
+    gif.id = 'wpstats';
+    console.log(gif);
```
jameswburke, line 26: ?
```suggestion
```

packages/integrations/components/integrationsManager/componentMap.js
```diff
@@ -1,13 +1,15 @@
 // Integration components.
 import CoralEmbed from '../coral';
 import YoastSchema from '../yoastSchema';
+import SiteStats from '../siteStats';
```
jameswburke, line 4: Can we call all of this `JetpackSiteStats` so it's clearer? Backend and frontend.

tylerreckart: 👍

tylerreckart: I'll include that update as part of my package changes for the `withPico` coral wrapper so that I don't have to switch contexts in my current work.

PR #272, opened 2020-08-14
IRV-643: allow embed to render from JSON template

tylerreckart:
See alleyinteractive/wp-irving#222 for relevant WP component files.


packages/integrations/components/coral/readme.md
```diff
@@ -0,0 +1,29 @@
+# Component for integrating Coral comments.
```
joemcgill, line 1: 💖 Docs!

PR #273, opened 2020-08-14
Styled component updates

ostowe:
N/A

* Modify the logo component to support children (used for an SVG logo in this case)
* Add `siteTheme` helper function

Tested on client project

### DIFF
--- a/packages/styled/utils/siteTheme.js
+++ b/packages/styled/utils/siteTheme.js
@@ -1,0 +1,7 @@
+import get from 'lodash/fp/get';
+
+const siteTheme = (valuePath) => (props) => (
+  get(`theme.${valuePath}`, props)
+);
+
+export default siteTheme;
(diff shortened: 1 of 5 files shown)


packages/styled/utils/siteTheme.js
```diff
@@ -0,0 +1,7 @@
+import get from 'lodash/fp/get';
+
+const siteTheme = (valuePath) => (props) => (
+  get(`theme.${valuePath}`, props)
+);
```
jameswburke, line 5: Can we add support for `lodash/get`'s [`defaultValue` parameter](https://lodash.com/docs/4.17.15#get)?

ostowe: mm, ok, `defaultValue` doesn't exist in the `fp` version of it, but we also don't really need to use the `fp` version.

jameswburke: What's fp?

PR #275, opened 2020-08-20
DEF-56: withPico component wrapper for Coral comments

tylerreckart:
__The core backend work required for this integration can be found at alleyinteractive/wp-irving#229__

## Summary
This PR introduces the `withPico` component wrapper for connecting a Coral comments instance to Pico via SSO. This integration allows users to login with their Pico account, which then acts as an identity provider to generate a corresponding JWT token that is used to either sign in or create a user account in Coral.

The `withPico` wrapper is not part of the export map for the integrations package due to its specific use-case. The only change needed to use Coral with Pico will be in your `componentMap.js` file. Example:
```js javascript
import ComponentMap from '@irvingjs/styled-components';
import userThemes from './themes';
import { IntegrationsManager } from '@irvingjs/integrations';
import CoralEmbed from '@irvingjs/integrations/components/coral/withPico';
export default {
  ...ComponentMap(userThemes),
  'irving/coral': CoralEmbed,
  'irving/integrations': IntegrationsManager,
};
```
The process for including Coral inside of a project's template JSON will remain unchanged when using the `withPico` component wrapper. 

### Other notable changes
The Pico integration has been updated to fix an issue where multiple `pico-widget-container` nodes would be appended to the DOM on each client-side render. The node is now only mounted on an initial server-side load.

Two additional DOM elements are now also being injected at the root of the body, `PicoSignal-container` and `PicoRule-button`. Both of these elements are injected outside of the context of Irving's root element in order to allow the application to validate their existence prior to firing the `pico.load` event. If those elements do not exist when that event is fired, Pico will not be able to detect their existence in the DOM.

The `PicoSignal-container` element serves as a store for the current user's status and information. Any time a user logs in, changes their information, payment status or logs out, this element will be mutated to reflect that change.
  - If you wish to take any actions based on changes to that element, you can mount a [mutation observer](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) to listen for changes and dispatch corresponding actions based on the values of those changes. See `components/coral/withPico.js` for an example of how to mount a mutation observer on the signal node.

The `PicoRule-button` element serves as a trigger to interact with the Pico widget programatically. You can retrieve the rule node with `document.getElementById('PicoRule-button')` and dispatch click events to summon the Pico widget without the user having to interact with Pico directly.


packages/integrations/components/pico/utils.js
```diff
@@ -0,0 +1,33 @@
+export default function mountPicoNodes(pageInfo) {
+  let nodesMounted = false;
+
+  // Mount the Signal and Rule nodes to be targeted by the Pico embed.
+  const documentBody = document.getElementsByTagName('body')[0];
```
xmattus, line 5: Are the elements added here defined by the Pico scripts? If so, is there a scenario we're correcting for here where they're not added by Pico's own embed? If not, is there a rationale for storing these values in document elements instead of the Redux store?

packages/integrations/components/coral/withPico.js
```diff
@@ -0,0 +1,107 @@
...
+import { useDispatch, useSelector } from 'react-redux';
+import CoralEmbed from './index';
+import {
+  actionVerifyPicoUser,
+  actionReceiveCoralLogoutRequest,
+} from '../../actions';
+import { tokenSelector } from '../../selectors/coralSelector';
+
+const withPico = (ChildComponent) => (props) => {
+  const [verificationRequestSent, setRequestStatus] = useState(false);
```
xmattus, line 11: I haven't used the new React hooks myself much, but I take it using them here is justifiable since it's simpler snd there's no real scenario where components elsewhere in the app would need to concurrently look at the Pico state? Would there be a scenario where you'd want this in the global Redux store instead of with `useState`?

packages/integrations/components/pico/utils.js
tylerreckart: @xmattus Great question! From my understanding of the [Pico documentation](https://help.trypico.com/en/articles/2906319-using-pico-signals) (which there isn't much of any it isn't really centered on a developer's perspective), they do not provide any container elements that are have the target `PicoSignal` class name. Pico uses this element as a sort of container for their widget to read and store user data based on the actions the user takes within their embed.

This element is being defined here outside of the root context of the Irving application so that it is available to Pico at all times. It gets injected on an initial server-load and is available through every subsequent client-side route change because it lives outside of Irving's root node. I had to go this route because Pico fires two events on every `visit`, `pico.init` and `pico.load`. If the target elements (both the signal and rule nodes) don't exist in the DOM by the time those actions are fired, Pico won't be able to find them and execute the appropriate behavior. It was possible to mount these elements within the context of the `withPico` HOC wrapper, but that ended up getting pretty messy and I had to resort to listening for `LOCATION_CHANGED` actions and delaying firing the `visit` event until I knew for a fact that the component had rendered.

As far as storing these values in the Redux store goes, it would definitely be doable, but I don't think it's necessary in this context. Since Pico targets any DOM element with the `PicoSignal` class and adds the user information as data attributes to that node automatically I don't really see a need for that information to exist in multiple places at once.

packages/integrations/components/coral/withPico.js
tylerreckart: @xmattus I don't believe so. The state value being set here is more of a flag to tell the component that a verification request has already been sent and to not execute another if the `MutationObserver` fires more than once, which I did experience during the development process.

packages/integrations/components/pico/utils.js
xmattus: Ah, ok. That makes sense, and I recall a bit of this from when I did the Pico implementation for the JFC 1.0 (née Civil First Fleet) sites.

packages/integrations/components/coral/withPico.js
xmattus: Sounds good to me!

packages/integrations/components/coral/withPico.js
```diff
@@ -0,0 +1,107 @@
...
+      const signalNode = document.getElementById('PicoSignal-container');
+      // Only mount the observer if the PicoSignal button exists in the DOM and
+      // the Pico data attributes have been appended to the element.
+      if (
+        signalNode &&
+        null !== signalNode.getAttribute('data-pico-email') &&
+        false === verificationRequestSent
+      ) {
+        // Define the observer.
+        const observer = new MutationObserver((mutations) => {
```
ostowe, line 42: Not necessary for this PR maybe, but I'd love to see an abstraction of a `MutationObserver` hook (maybe `useMutation`?) in irving core. We have one for intersection observers currently.

packages/integrations/components/pico/utils.js
```diff
@@ -0,0 +1,33 @@
...
+  // Append a container node for the Pico signals to target into the DOM if one doesn't already exist.
+  if (documentBody && ! document.getElementById('PicoSignal-container')) {
+    // Create the signal node.
+    const signalNode = document.createElement('div');
+    // Set the attributes.
+    signalNode.setAttribute('id', 'PicoSignal-container');
+    signalNode.setAttribute('class', 'PicoSignal');
+    signalNode.setAttribute('style', 'display: none');
+    // Append the node into the DOM.
+    documentBody.appendChild(signalNode);
```
ostowe, line 15: Is this maybe an excuse to use [portals](https://reactjs.org/docs/portals.html#usage)?  So like:
```javascript
const SignalNode = () => (
  <div
    id="PicoSignal-container"
    class="PicoSignal"
    style={{ display: 'none' }}
  />
);

React.createPortal(SignalNode, document.getElementById('PicoSignal-container'));
```

tylerreckart: @ostowe Great idea. I tried implementing that but could not get the nodes to mount in a reliable way for all renders. I think since these components live outside of the context of React and are only used for interacting/tracking actions in the Pico widget embed that the plain JS nodes should be sufficient here.

packages/integrations/components/coral/withPico.js
tylerreckart: I'm definitely for that. I'll put together an integration ticket for that work.

PR #277, opened 2020-08-24
Add Coral Comment Count component

joemcgill:

DEF-111

Adds:
* CommentCount component inside the Coral integrations folder for displaying comment counts of an article from Coral.

This can be tested along with https://github.com/alleyinteractive/wp-irving/pull/232 or manually, by hard coding a component response in a layout file that passes the following values:

* `articleUrl` – The URL of an article.
* `embedUrl` – The URL to a Coral instance.
* `noText` – Boolean (optional), whether to hide the text and just show the number of comments.

### DIFF
--- a/packages/integrations/components/coral/commentCount.js
+++ b/packages/integrations/components/coral/commentCount.js
@@ -9,7 +9,8 @@ const CommentCount = (props) => {
     noText,
   } = props;
 
-  if (! embedUrl) {
+  // We need an embedUrl and articleUrl to proceed.
+  if (! embedUrl || ! articleUrl) {
     return null;
   }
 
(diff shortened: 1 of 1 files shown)


packages/integrations/components/coral/commentCount.js
```diff
@@ -0,0 +1,53 @@
...
+        defer: true,
+      },
+    }
+  );
+
+  useEffect(() => {
+    if (window.Coral) {
+      console.log('Coral Comment Count');
+    }
+  }, [loaded]);
```
jameswburke, line 32: Is this still needed?

joemcgill: Not really. Removed in 8767b60.

PR #283, opened 2020-08-28
DEF-193: Extend Pico/Coral integration components for custom usernames

tylerreckart:
This PR accomplishes the AC as defined in DEF-193 and DEF-194. The full integration is still a work in progress.


packages/integrations/components/coral/usePicoObserver.js
```diff
@@ -0,0 +1,138 @@
...
+
+/**
+ * A hook that mounts a MutationObserver on the global `PicoSignal-container`
+ * DOM node on the `pico.loaded` event. The observer dispatches actions used
+ * to verify that user with Pico and perform conditional rendering/SSO-token
+ * retrieval given the values of the Signal node's data attributes.
+ *
+ * @param {array} tiers Array of valid subscription tier IDs for Coral SSO.
+ */
+export default function usePicoObserver(tiers) {
```
tylerreckart, line 22: I need to do a whole round of clean up on this hook to make sure everything is working exactly as it should and will be writing unit tests for it in a future PR. This part of the integration is still in progress, but should not block the review of this work.

packages/integrations/components/coral/usePicoObserver.js
```diff
@@ -0,0 +1,138 @@
...
+                  // Dispatch an action that signifies the user has upgraded
+                  // their subscription to a level sufficient to enable Coral SSO.
+                  if (showUpgradeModal) {
+                    dispatchReceivePlanUpgrade();
+                  }
+
+                  if (! requestSent) {
+                    // Set a delay in order for the global Pico object to update with
+                    // the user's ID prior to dispatching the verification request.
+                    setTimeout(() => {
```
joemcgill, line 102: Oh, fun race conditions! Since we don't care about IE11, I wonder if this would be a good use case for using a [JS Proxy object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy)?

joemcgill (approved):
All looks good to me. ⚡

packages/integrations/components/coral/usePicoObserver.js
tylerreckart: Interesting thought. I'll give that implementation some experimentation time in DEF-235

PR #284, opened 2020-08-29
Add proptypes and some transform logic for standardized component props

ostowe:
[**IRV-652**](https://alleyinteractive.atlassian.net/browse/IRV-652)

* Add a hook for getting and transforming a specific set of standard props:
  * `className`, which will be combined together using the `classnames` package
  * `id`, which will render directly to the `id` attribute
  * `style`, in which any site theme context variables will be replaced
  * `tag`, which will be converted to the styled-components `as` prop (or omitted entirely if no tag is provided)
  * All of the above can have component defaults which will be combined with user input or, in the case of `id`, replaced by user input.
* Add a utility for converting a component name to a className. This className will automatically be added to a component.
* `PropTypes` definitions for a wider variety of standard props.
* `defaultProps` definitions for the same.
* Fix some mismatched packages (identified by `npm run develop:bootstrap`)

* I've written unit tests for the `useStandardProps` hook and surrounding utilities
* I've tested in a client project.


packages/styled-components/components/fragment/index.js
```diff
@@ -12,22 +15,18 @@ const Fragment = (props) => {
...
-
-  // If we have something in `style` and element is empty, require a tag.
-  if (null === tag && Object.keys(style).length && ! style.length) {
-    tag = 'span';
-  }
+  const standardProps = useStandardProps(props);
+  const { as } = standardProps;
+  const useTag = (
+    as ||
+    (! as && Object.keys(style).length && ! style.length)
```
joemcgill, line 23: The `! as` in the second grouping is unnecessary, since if it is a truthy value, this will never be evaluated.

packages/styled-components/package.json
```diff
@@ -29,15 +29,15 @@
...
+    "react-hook-form": "^6.0.6",
     "react-icons": "^3.10.0",
     "styled-css-grid": "^1.2.1"
   },
   "devDependencies": {
     "@irvingjs/babel-preset-irving": "^3.3.0-alpha.1",
     "@irvingjs/core": "^3.3.0-alpha.1",
-    "@testing-library/jest-dom": "^5.11.0",
-    "@testing-library/react": "^10.4.5",
+		"@testing-library/jest-dom": "^5.11.4",
```
joemcgill, line 39: More weird whitespacing.

packages/styled/hooks/useStandardProps.js
```diff
@@ -0,0 +1,90 @@
...
+ * Provide a component with a standardized set of props for that component to use.
+ * @param {object} props    React comonent props.
+ * @param {object} componentDefaults Overrides for default prop values.
+ */
+const useStandardProps = (props, componentDefaults = {}) => {
+  const {
+    className = '',
+    id = '',
+    style = {},
+    tag = '',
```
joemcgill, line 57: I'm unsure if `tag` really belongs here. I would assume that `standardProps` should be able to be applied to most, if not all components. Can you pass a tag to most of our core components in order to transform the generated markup?

packages/styled/hooks/useStandardProps.js
```diff
@@ -0,0 +1,90 @@
...
+    componentName,
+  } = props;
+  const componentNameClass = componentName ?
+    createComponentNameClass(componentName) : '';
+  // Using the `styled-components` ThemeContext, attempt to use dynamic
+  // values in CSS properties.
+  const siteTheme = useContext(ThemeContext);
+  const standardizedProps = {
+    className: classNames(
```
joemcgill, line 66: I love this update. So smooth.

packages/styled/hooks/useStandardProps.test.js
```diff
@@ -0,0 +1,90 @@
+import { renderHook, act } from '@testing-library/react-hooks';
+import SiteThemeProvider from '../components/siteThemeProvider';
+import useStandardProps, {
+  replaceWithSiteTheme,
+  createComponentNameClass,
+} from './useStandardProps';
+
+/* eslint-disable max-len */
+describe('useStandardProps', () => {
```
joemcgill, line 9: 💖 Tests!

packages/styled-components/components/bodyWrapper/index.js
```diff
@@ -3,54 +3,63 @@ import PropTypes from 'prop-types';
...
 
 const BodyWrapper = (props) => {
-  const { bodyClasses, children, theme } = props;
+  const {
+    bodyClasses,
+    children,
+    theme,
+  } = props;
+  const standardProps = useStandardProps(props, {
+    id: 'content',
```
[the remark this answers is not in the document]
ostowe: Hmm, that existed primarily to support a built-in "Skip to content" link in the `App` component, but that link appears to no longer exist so maybe we can remove this.

packages/styled/hooks/useStandardProps.js
ostowe: I'm a bit on the fence here, but I lean toward leaving it in because this is in the `styled` package, which assumes you're planning to primarily use styled components for your styling.

PR #287, opened 2020-08-31
feat(image): update component to handle responsive images

joemcgill:
This adds support for srcset, sizes, height, and width attributes to help performance and to reduce
layout shifts during lazy loading.

BREAKING CHANGE: Removes unused support for photon transforms and picture element sources.


packages/styled-components/components/image/index.js
```diff
@@ -36,16 +36,16 @@ const Image = (props) => {
     className,
     credit,
     fallbackSrc,
+    height,
     loading,
     objectFit,
-    // eslint-disable-next-line no-unused-vars
```
joemcgill, line 41: I went ahead and removed these since they're currently unused and I think we can better implement them elsewhere.

PR #290, opened 2020-09-02
fix(byline): remove errant span tag in component with 2 children

amyevans:

N/A

Remove the `<span>` that exists only in the markup for the case of 2 children, which is inconsistent from the markup produced by the other cases (1 child, 3+ children)

Tested locally


jameswburke (approved):
:bee:

PR #292, opened 2020-09-03
DEF-235: Refactor MutationObserver pattern for Pico/Coral integrations

tylerreckart:
This PR cleans up the Pico `MutationObserver` pattern introduced in alleyinteractive/irving#283. Among the changes are:
- A new hook in the integrations package, `useMutationObserver`. This hook allows us to mount a mutation observer inside of any React component given a `ref`, a callback, and an optional observer configuration. @jameswburke I'm flagging this for you as I think this hook would be a good candidate for moving into core in the future, but right now is only utilized inside of the Integrations package as part of the Pico/Coral integration.
- Refactors the `usePicoObserver` hook into a React component that uses the `useMutationObserver` hook to manage local state within the component and dispatch actions based on the user state accordingly.
- General clean up around action/reducer logic.


packages/integrations/components/pico/observer.js
```diff
@@ -0,0 +1,200 @@
...
+
+    const hasValidStatus = 'registered' === status || 'paying' === status;
+    const hasNotUpdated = ! isAnonymous || ! isRegistered || ! isPaying;
+
+    if ('excluded' !== status && hasValidStatus && hasNotUpdated) {
+      // Update the `email` value with the user's email address.
+      if (undefined === email) {
+        setEmail(emailAddress);
+      }
+      // If the user's tier is included in the whitelisted tiers for SSO,
```
joemcgill, line 46: Avoid use of "whitelist" in favor of more inclusive language.

```suggestion
      // If the user's tier is included in the allowed list of tiers for SSO,
```

packages/integrations/components/pico/observer.js
```diff
@@ -0,0 +1,200 @@
...
+      }
+      // If the status is paying, set the isPaying flag to true.
+      if ('paying' === status) {
+        setIsPaying(true);
+      }
+    }
+
+    // If the user was previously recognized as registered or paying, but
+    // is then anonymized, update the state values accordingly.
+    if ((isRegistered || isPaying) && 'anonymous' === status) {
```
joemcgill, line 63: Is there a case where `isPaying` would be true, but `isRegistered` would not be?

packages/integrations/components/pico/observer.js
```diff
@@ -0,0 +1,200 @@
...
+
+  const observerCallback = debounce((mutations) => {
+    mutations.forEach((mutation) => {
+      const { target, type, attributeName } = mutation;
+
+      if ('attributes' === type && 'data-pico-status' === attributeName) {
+        manageLocalState(target);
+      }
+    });
+  }, 250);
```
joemcgill, line 87: We may want to consider a smaller delay here if this introduces a perceptible delay in any UI changes after login.

packages/integrations/components/pico/observer.js
```diff
@@ -0,0 +1,200 @@
...
+
+  // Declare a ref for the `useMutationObserver` hook to watch.
+  const signalRef = useRef();
+  // Mount the MutationObserver.
+  useMutationObserver(signalRef, observerCallback);
```
joemcgill, line 92: This is a really nice pattern. 💖

packages/integrations/components/pico/observer.js
```diff
@@ -0,0 +1,200 @@
...
+      return;
+    }
+
+    if (! accessToken) {
+      // If the user is registered but non-paying and the upgrade modal
+      // has not been show, update the store to trigger the modal.
+      if (
+        (
+          isRegistered ||
+          isPaying
```
joemcgill, line 139: If I'm reading your inline comment correctly, should this be `! isPaying`?

joemcgill (approved):
A few questions about some of the payment logic, but overall this looks like really nice work. 👍

packages/integrations/components/pico/observer.js
tylerreckart: That comment was incorrect. Updating 😄

packages/integrations/components/pico/observer.js
tylerreckart: Yes. Pico uses `registered` and `paying` as separate identifiers in the `data-pico-status` attribute. If a user is registered with Pico, but has not subscribed, they will be marked as registered. If they are paying, regardless of tier, they'll simply be marked as paying.

PR #293, opened 2020-09-03
Fix the query string handling for the pagination component

jameswburke:
None.

Ensure that query vars are maintained while paginating.

Tested locally.


mslinnea (approved):
🐴

PR #296, opened 2020-09-03
feat(img): Add tests for the styled image component

joemcgill:
This adds an initial set of tests for the image component to ensure that
properties are being set correctly in the DOM in various combinations.

This also fixes a bug where an image component was rendered even when there
was no `src` or `fallbackSrc` properties present.


packages/styled-components/components/image/image.test.js
```diff
@@ -0,0 +1,68 @@
...
+    const testProps = {
+      src: 'https://test.com/test.jpg',
+      caption: 'Test caption',
+      credit: 'Example credit',
+      showMeta: false,
+    }
+
+    const { container, queryByText } = render(<Image { ...testProps } />);
+
+    // Ensure caption and credit are rendered.
```
renatonascalves, line 64:
```suggestion
    // Ensure caption and credit are not rendered.
```

renatonascalves (approved):
💯

PR #297, opened 2020-09-04
fix(purgecache): remove duplicative res.end()

amyevans:

N/A

* Remove `res.end()` call which was being called too early and therefore resulting in the following error:
```
Error [ERR_STREAM_WRITE_AFTER_END]: write after end
```
(which in practice was causing the Redis cache to not clear)

Tested locally


jameswburke (approved):
:bee:

PR #298, opened 2020-09-04
Separate component responsible for rendering roots from app wrapper

ostowe:
* Split up app component and component responsible for rendering connected roots to prevent unnecessary rerenders.
* Add `h1` and `time` tags to html `allowedTags` because gutenberg supports them.

Tested in client project.

### DIFF
--- a/packages/core/components/root/index.js
+++ b/packages/core/components/root/index.js
@@ -35,7 +35,6 @@ Root.propTypes = {
 const mapStateToProps = (state) => ({
   roots: getRoots(state),
   providers: getProviders(state),
-  error: !! state.error,
 });
 
 export default connect(mapStateToProps)(Root);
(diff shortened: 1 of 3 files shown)


packages/core/components/root/index.js
```diff
@@ -0,0 +1,41 @@
...
+  /**
+   * Root provider configurations
+   */
+  providers: PropTypes.arrayOf(PropTypes.object).isRequired,
+};
+
+const mapStateToProps = (state) => ({
+  roots: getRoots(state),
+  providers: getProviders(state),
+  error: !! state.error,
```
joemcgill, line 38: Do we need this duplicate error prop mapping in both places, or can it be passed from the `App` container?

joemcgill (approved):
Looks good to me. 👍

PR #299, opened 2020-09-05
fix(adminbar): fix React error regarding a change in the order of hooks

amyevans:

N/A

Fixes the following warning:
```
React has detected a change in the order of Hooks called by AdminBar.
```
This error was happening because the `useEffect` hook was not present for all renders

Tested locally


davisshaver (approved):
🔨

PR #303, opened 2020-09-08
fix(integrationsManager): Remove Redux caching of integrations data

joemcgill:

This removes the mechanism that caches integrations data to the Redux store, and instead prefers to
re-render fresh integration data each time the integrations manger is rendered and recieves updates
integration data. This fixes an issue where fresh data being passed from API responses were not
being passed as props to components during page renders.

After applying this change, you can see that the GTM integration and Pico integration is rendering with fresh data from the components API response.

### DIFF
--- a/packages/integrations/components/integrationsManager/index.js
+++ b/packages/integrations/components/integrationsManager/index.js
@@ -2,9 +2,6 @@ import React from 'react';
 import PropTypes from 'prop-types';
 // User Config.
 import userConfig from '@irvingjs/irving.config';
-// Redux.
-import { connect } from 'react-redux';
-import { actionHydrateComponents } from '../../actions';
 // Component Map.
 import componentMap from './componentMap';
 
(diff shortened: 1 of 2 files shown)


packages/integrations/components/integrationsManager/index.js
```diff
@@ -2,62 +2,52 @@ import React from 'react';
...
-        if ('integrations' in userConfig && key in userConfig.integrations) {
-          props = userConfig.integrations[key];
-        } else {
-          props = integrations[key];
-        }
-        // Update the item at the current index with the transformed component.
-        components.push({ key, props });
-      }
-    });
+  componentMap.forEach((component) => {
```
joemcgill, line 34: This can probably further be refactored to avoid the need for this mutated integrations array, but was trying to avoid too many logic changes so this doesn't introduce regressions.

packages/integrations/components/integrationsManager/index.js
```diff
@@ -95,48 +85,10 @@ const IntegrationsManager = ({
...
-  hydratedComponents: [],
 };
 
 IntegrationsManager.propTypes = {
-  integrations: PropTypes.shape({
-    googleAnalytics: PropTypes.shape({
-      trackingId: PropTypes.string,
-    }),
-  }),
-  isHydrated: PropTypes.bool.isRequired,
```
ostowe, line 107: So this will just remove all the caching logic entirely? I thought we discussed the possibility of bypassing it only for certain components. I'm not opposed to this approach if it works, though maybe worth revisiting later.

packages/integrations/components/integrationsManager/index.js
```diff
@@ -2,62 +2,52 @@ import React from 'react';
...
-  hydrateComponents,
-  hydratedComponents,
-}) => {
-  // This check is run on server-side renders to ensure that the component hydration
-  // is ran a single time. After hydration, `isHydrated` is set to `true` and will
-  // only be re-run if the component props change during the component's lifecycle.
-  if (false === isHydrated) {
-    // Retrieve the keys available for render in the component map.
-    const keyMap = Object.keys(integrations);
+const IntegrationsManager = ({ integrations }) => {
```
ostowe, line 17: More of a pet peeve than something wrong here, but I generally prefer to destructure in the function body instead of params to allow for easier debugging and/or console logging. I think we've done this differently in different projects, but I'd prefer to standardize around not use destructuring in params.

ostowe (approved):
lgtm 💯

packages/integrations/components/integrationsManager/index.js
joemcgill: I definitely think we should revisit later, but the current caching mechanism seem to not be working at all, so this just removes all of that logic entirely.

packages/integrations/components/integrationsManager/index.js
joemcgill: I don't have strong opinions about which pattern we use. This component was already destructuring in params, so I left it.

packages/integrations/components/integrationsManager/index.js
```diff
@@ -2,62 +2,52 @@ import React from 'react';
...
+  // Test for empty objects.
+  const isEmpty = (obj) => (
+    0 === Object.keys(obj).length && obj.constructor === Object
+  );
+
+  // Bail early if no integrations exist.
+  if (isEmpty(integrations)) {
```
tylerreckart, line 24: 👏

tylerreckart (approved):
🧪

PR #308, opened 2020-09-09
Convert Play/Pause button component to a themeable styled component

ostowe:
[**IRV-670**](https://alleyinteractive.atlassian.net/browse/IRV-670)

Allow users to theme the play/pause button in the audio player package more easily

Tested in client project


packages/audio-player/package.json
```diff
@@ -22,6 +22,7 @@
   "author": "Alley (Alley.co)",
   "license": "MIT",
   "dependencies": {
+    "@irvingjs/styled": "^4.0.0-beta.2",
```
jameswburke, line 25: Should this be committed to the plugin? Should we use a stable release?

jameswburke (approved):
:bee:

PR #311, opened 2020-09-09
DEF-36: Add GTM event attributes

davisshaver:
Adds data attributes that can be listened for using Google Tag Manager.

### DIFF
--- a/packages/styled-components/components/searchForm/index.js
+++ b/packages/styled-components/components/searchForm/index.js
@@ -57,6 +57,7 @@ const SearchForm = (props) => {
         ref={register}
         type="text"
       />
+      {/* @TODO Add GTM event.  */}
       <SearchFormSubmitButton type="submit">Search</SearchFormSubmitButton>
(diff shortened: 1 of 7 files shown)


packages/styled-components/components/searchForm/index.js
```diff
@@ -57,6 +57,7 @@ const SearchForm = (props) => {
         ref={register}
         type="text"
       />
+      {/* @TODO Add GTM event.  */}
```
joemcgill, line 60: Did you mean to do this before this was merged?

joemcgill (approved):
Looks good to me. In the future we should probably come up with a more elegant way to add data attributes to components but this gets what we need for now.

packages/styled-components/components/searchForm/index.js
davisshaver: @joemcgill No, but I'll make a new ticket so it doesn't get lost in the shuffle.

PR #315, opened 2020-09-09
fix(pico): Fix pageview data not being updated properly

joemcgill:

DEF-289

This fixes an issue where visits were being triggered before new Pico data had been updated and
ensures that we always pass a true article value whenever we are on a post.


ostowe (approved):
💯

PR #316, opened 2020-09-09
hotfix/fix pico pageinfo updates

joemcgill:
This reverts a change that is causing the initial container not to load.


tylerreckart (approved):
👍

PR #321, opened 2020-09-09
DEF-36: Add additional GTM events

davisshaver:
Adds GTM attributes for logo link and search.

### DIFF
--- a/packages/styled-components/components/logo/index.js
+++ b/packages/styled-components/components/logo/index.js
@@ -36,6 +36,9 @@ const Logo = (props) => {
     <LogoWrapper {...standardProps}>
       <LogoLink
         as={Link}
+        gtmAction="Homepage"
+        gtmCategory="Navigation"
+        gtmLabel="Logo"
         href={href}
       >
         {(logoImageUrl && ! hasChildren) ? (
--- a/packages/styled-components/components/searchForm/index.js
+++ b/packages/styled-components/components/searchForm/index.js
@@ -52,12 +52,13 @@ const SearchForm = (props) => {
       onSubmit={handleSubmit(onSubmit)}
     >
       <SearchFormTerm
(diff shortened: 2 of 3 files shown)


packages/styled-components/components/logo/index.js
```diff
@@ -36,6 +36,9 @@ const Logo = (props) => {
     <LogoWrapper {...standardProps}>
       <LogoLink
         as={Link}
+        gtmAction="Homepage"
```
joemcgill, line 39: Are we sure this should always be "Homepage?"

joemcgill (approved):
I'd like to see us make these configurable in case a project wants to use different GTM values, but this works for now. 🏷️

packages/styled-components/components/logo/index.js
davisshaver: @joemcgill I guess not – maybe we should move these to the `irving/site-link` definition in WP Irving? What do you think?

jameswburke: I'd like us to consider making `gtm` an object with each of these properties, and I also agree that setting the defaults in WP Irving is ideal.

davisshaver: @jameswburke @joemcgill Are you two open to getting this out as-is and then I will follow up to make any GTM values in WP Irving or Irving configurable from WP Irving?

Other question: Do we have any prior art for this? Or suggestions on existing features to look at for inspiration on approach?

jameswburke: @davisshaver, I confirmed with @joemcgill that we'd like to see this updated since it's not blocking anything and we need to go through the release process. We can both swarm with you today to help get it over the finish line.

joemcgill:
Closing as this has been superseded by #372

PR #323, opened 2020-09-10
feat(socialsharing): update Social Sharing component to allow more flexibility

amyevans:

N/A

This change will provide a little more flexibility for defining properties of individual social share platforms, e.g. setting the `target` per platform. It could also be extended more easily later to allow defining of the icon name, icon title, etc.

Tested locally


tylerreckart (approved):
😲

PR #325, opened 2020-09-10
hotfix(pico/coral): explicitly summon ugrade modal on Coral UI button click for registered and non-allow-listed users

tylerreckart:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,14 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+# [4.1.0-alpha.1](https://github.com/alleyinteractive/irving/compare/v4.1.0-alpha.0...v4.1.0-alpha.1) (2020-09-10)
+
+**Note:** Version bump only for package irving
+
+
+
+
+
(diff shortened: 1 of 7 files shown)


xmattus (approved):
This is OK for now, but we're imposing some non-general business logic on how the Pico tier is used — it's not universally the case that Coral login via Pico would only come with paid but not registered status. We should eventually expand this to take its behavior from kind of configuration or other indicator of what Pico status confers Coral SSO eligibility.

tylerreckart:
> This is OK for now, but we're imposing some non-general business logic on how the Pico tier is used — it's not universally the case that Coral login via Pico would only come with paid but not registered status. We should eventually expand this to take its behavior from kind of configuration or other indicator of what Pico status confers Coral SSO eligibility.

@xmattus The logic here is using a configuration sent from WP for allow-listed commenting tiers for coral SSO. If the current user is paying but their tier is not explicitly set in the allowed tiers in WP, the modal logic will be triggered. I believe this satisfies what you are positing, but I could be wrong.

xmattus:
@tylerreckart Ah, ok, I didn't realize it was configurable to that extent in WP. Disregard, then, since I suppose we have to pick some standard default in the code.

PR #326, opened 2020-09-10
Update logic for providers to cache data more effectively

ostowe:
[**IRV-675**](https://alleyinteractive.atlassian.net/browse/IRV-675)

* Update redux state for providers to allow storing data under arbitrary keys, updated only if necessary.
* Simplify logic for generating and nesting providers
* Allow providers to _just_ exist in redux if necessary, letting components down the tree access the provider's data via a selector instead of React context.

* I've written unit tests for all functionality


packages/core/reducers/providersReducer.js
```diff
@@ -0,0 +1,86 @@
...
+  const { providers } = payload;
+  const newProviders = providers.reduce((acc, provider) => {
+    const newProvider = provider;
+    const {
+      name,
+      config: { providerKey },
+    } = newProvider;
+
+    // If user specifies 'route' as they provider data key,
+    // key new data for every route change.
```
renatonascalves, line 69:
```suggestion
    // If a user specifies 'route' as their provider data key,
    // key new data for every route change.
```

renatonascalves (approved):
💯

PR #332, opened 2020-09-14
feat: added a section on theming styled-components to Storybook

techmeg:
This section only addresses theming styled-components. I would guess we should add to it to address wp-components and maybe the audio-player component? Happy to keep going after this section is reviewed.


stories/02-features/05-Style-with-themes.stories.mdx
```diff
@@ -0,0 +1,71 @@
...
+export const ContainerWrapper = styled.div`
+  margin: auto;
+  width: 100%;
+`;
+/* eslint-enable */
+
+```
+
+The conditions needed to apply themes are:
+* a `themes` folder must be created in the irving directory with an `index.js` file within. It's a good practice to organize your themes by component name, so to style a container, for example, create a `container` folder within the `themes` folder.
```
lmfern, line 20: I think it might help to include the path to the irving directory here, and err on the side of overcommunication. I also really love that you've provided an example of the folder structure below!

stories/02-features/05-Style-with-themes.stories.mdx
```diff
@@ -0,0 +1,71 @@
...
+* a JavaScript file with your custom styling must to imported to the `index.js` file.
```
lmfern, line 21:
```suggestion
* a JavaScript file with your custom styling must be imported to the `index.js` file.
```

lmfern (approved):
👍 These are clear and thorough directions for theming styled components. I especially appreciate the example and explanation of each step therein!

A thought related to theming - somewhere, either on this page or an adjacent section of Storybook, it  could be helpful for us to explain “how to know what themes are available for each component" ie  the themes that ship with each Irving component (and a link or path to where a person can view those) as well as the custom themes for that component (as is described here in this Storybook entry, located in `client/irving/themes/<name of component>`)

stories/02-features/05-Style-with-themes.stories.mdx
```diff
@@ -0,0 +1,71 @@
+<Meta title="Features|Style with Themes" />
+
+# Theming a Styled Component
+
+An assortment of minimally Styled Components is included in Irving' styled-components package. Styles are easily overwritten by providing Irving with your own custom themes. Each styled-component's configuration includes a theme property that maps to the default style included in the component package. For example, the default.js theme for the container component is:
```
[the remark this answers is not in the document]
ostowe: Agreed on "what is a theme" section, perhaps describing a bit how the `withThemes` higher-order component works and why we built it the way we did (I'm happy to coordinate with you on this)

stories/02-features/05-Style-with-themes.stories.mdx
```diff
@@ -0,0 +1,71 @@
...
+```
+
+The conditions needed to apply themes are:
+* a `themes` folder must be created in the irving directory with an `index.js` file within. It's a good practice to organize your themes by component name, so to style a container, for example, create a `container` folder within the `themes` folder.
+* a JavaScript file with your custom styling must to imported to the `index.js` file.
+* the styled-component in your template must point to your custom-theme.
+
+The styled-components's `default.js` is the best place to start when writing your own theme so copy it from `node_modules/@irvingjs/styled-components/components/<component-to-theme>/themes`. From there, add or subtract any styles you like. You can overwrite the default behavior by naming your custom theme `default.js` or keep the default behavior and add as many theme variants as you need.
+
+For example, you might want a flex container around a section, that is 90% wide with a yellow background. In order to point your app to your custom theme, you would apply the "theme" prop to your component:
```
ostowe, line 39: We shouldn't provide examples using `json` templates in here as that is a feature specific to `wp-irving`. I would say any docs that explicitly live outside the WordPress section should avoid referencing WordPress-specific functionality.

Instead, illustrate how passing a `themeName` prop to a styled component will determine which theme is rendered.

stories/02-features/05-Style-with-themes.stories.mdx
techmeg: check the added copy

stories/02-features/05-Style-with-themes.stories.mdx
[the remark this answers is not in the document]
techmeg: added intro and made suggested changes

stories/02-features/05-Style-with-themes.stories.mdx
```diff
@@ -0,0 +1,71 @@
...
+```js
+import styled from 'styled-components';
+
+/* eslint-disable import/prefer-default-export */
+export const ContainerWrapper = styled.div`
+  margin: auto;
+  width: 100%;
+`;
+/* eslint-enable */
+
```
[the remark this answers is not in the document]
techmeg: VS Code did not like it when I took out that extra line -- maybe because it's JS code? so I left it in

ostowe (commented):
This looks awesome @techmeg, thanks for writing this up! I left a couple comments for new sections, which I'm happy to co-write with you if that would be helpful.

stories/02-features/05-Style-with-themes.stories.mdx
lmfern: nice addition!

stories/02-features/05-Style-with-themes.stories.mdx
```diff
@@ -0,0 +1,216 @@
...
+# Styling with Themes
+Themes are one of a variety of ways that styling can be applied in Irving. Themes can be used in conjunction with the `@irvingjs/styled-components` package or with Irving's core styling setup as an elegant and efficient way to include your styles. It's also possible to style using a `className` or a `style` prop.
+
+## So, what's a theme?
+A theme in the Irving sense can be one of two things, depending on whether or not you're using the `styled-components` package:
+* Using `styled-components`, themes are a method of replacing one or more styled-components in an Irving component's JSX with your own.
+* Using Irving core alone, themes are a method of replacing one or more classnames in a component's stylesheet with your own.
+
+Irving's theming functionality may not work quite the way you expect from other themeing methods you have used in the past, and should not be confused with the [theme functionality built in to Styled Components](https://styled-components.com/docs/advanced#theming). Styled Components' built-in theming can be used as well, but is entirely separate from Irving's themes.
```
jameswburke, line 11: This disclaimer reaffirms my belief that we need a new name for this.

stories/02-features/05-Style-with-themes.stories.mdx
```diff
@@ -0,0 +1,216 @@
...
+</div>
+```
+
+That's it! As you can see, theming with Irving Core is relatively similar to theming with `styled-components`. Happy styling!
+
+## Additional notes on providing themes via WP-Irving
+
+Extending the styled-component theming plan, themes can also be applied to components in [WP Irving](https://github.com/alleyinteractive/wp-irving), the WordPress plug in to Irving. Components like post-list, site-menu, social sharing, etc., can be themed by determining the styled-component **alias** for that component and theming that component.
+
+For example `irving/post-title` is an alias for the `irving/text` component and can be themed via that component. You may apply a theme that already exists, like `h1`, or one you make a custom theme like `special-post-title`. If it's not obvious what the alias might be, use the WordPress editor to check the component's API and you'll see something like this:
```
jameswburke, line 197: > If it's not obvious what the alias might be, use the WordPress editor to check the component's API and you'll see something like this:

We can also point to http://storybook.irvingjs.com/?path=/docs/wordpress-components-table-of-contents--page

Each of those components documents what the alias is

stories/02-features/05-Style-with-themes.stories.mdx
```diff
@@ -0,0 +1,216 @@
...
+```
+
+As illustrated above, the `withThemes` HOC requires only a theme mapping object in which keys are a theme name and values are objects containing theme components. This HOC will provide a `theme` prop to the component to which the HOC is applied. This prop is an object on which both keys and values are the same (the name of each component function).
+
+As you may have noted already, the `withThemes` prop for styled components doesn't actually _require_ or integrate directly with styled components at all—it merely expects each value of the theme object to be a React component. This means that, for most intents and purposes, you will probably just supply an object of styled components in your theme. However, you can get creative if you want to (and at your own risk) by supplying any React component at all and, as long as the component is named the same as one of the theme values used in the themed component itself, it will be applied as-is. Be sure to render `children` as necessary, otherwise you may end up with chunks of your component that don't render (unless this is what you want!).
+
+### Theming components using Irving Core's CSS
```
jameswburke, line 142: Not right now, but I think we need to break this out into its own story in a new section for just styling.

jameswburke (approved):
:bee: some additional thoughts

stories/02-features/05-Style-with-themes.stories.mdx
ostowe: Ah yeah, this is better

stories/02-features/05-Style-with-themes.stories.mdx
ostowe: Agreed, I think I'm going to make four sections: styling with CSS, theming with CSS, styling with styled components, theming with styled components.

PR #333, opened 2020-09-15
DISCO-68 Accessibility Improvements

cahdeemer:
[DISCO-68](https://alleyinteractive.atlassian.net/browse/DISCO-68)

* Add a "skip to content" link 
* Replace the redundant aria landmark from the bodyWrapper component with an id of content (works with the skip link)
* Add an aria-label to the Menu component to help screen reader users quickly distinguish between nav landmarks
* Add a visually hidden label to the search form to help screen reader users understand the purpose of the input (not all screen readers will read the placeholder text)

I included a unit test and I tested this code with a project.


packages/styled-components/components/searchForm/index.js
```diff
@@ -51,7 +52,13 @@ const SearchForm = (props) => {
       action={baseUrl}
       onSubmit={handleSubmit(onSubmit)}
     >
+      <SearchLabel
+        id="search-label"
+      >
+        Search input
+      </SearchLabel>
```
isGabe, line 59: I noticed in the JFC repo that eslint wanted me to wrap the label around the input. This error would not go away until I did: `A form label must be associated with a control  jsx-a11y/label-has-associated-control`. 

Just a note, maybe for a future conversation with how we set up that rule. I don't think that should be forced, but I did it anyway there to make eslint shut up.

isGabe (approved):
💯

packages/styled-components/components/searchForm/index.js
cahdeemer: gotta keep eslint happy! made that change

PR #335, opened 2020-09-16
Update/add pico staging support

joemcgill:

DEF-323

This uses a `widgetUrl` prop passed from the component API as the basis for loading the widget script
so that we can swap between production and staging.

You can check the "use staging" option in the WP Irving plugin to pass the staging URL as part of the API response and Irving will load the pico widget code from the staging URL.


jameswburke (approved):
:bee:

PR #338, opened 2020-09-16
fix(pico): fix eslint error: single quotes

joemcgill:
Unsure why this didn't get flagged in the #335 PR, but it might break builds where single quotes are enforced.


ostowe (approved):
💯  strange, not sure why that wouldn't be caught either!

PR #339, opened 2020-09-17
fix(pico): Trigger Pico actions on pre-fetched pages

joemcgill:

IRV-710

This fixes a bug where pre-fetched pages were not triggering Pico actions
because the Pico component was being rendered from the previous page's
state. We had previously included a check for whether we had already
dispatched a Pico page update using a `useState` hook to bypass duplicate
dispatches and that state was not being reset between pageviews if the Pico
component instance was being reused.

This refactors the logic for side effects into several separate `useEffect`
hooks to separate responsibilities for each side effect and simplify the
number of dependencies each `useEffect` checks before running.


ostowe (approved):
💯 🥳

PR #343, opened 2020-09-18
feature/IRV-632 document use of inline style prop

techmeg:
Documenting use of inline styling in WordPress templates. Should be refined and moved to logical section -- in features section just as placeholder.

Tests: I know this code works because I tested locally.


lmfern (approved):
👍  This looks good to me. The note about `grid` and `cell` is particularly useful info right now.

PR #344, opened 2020-09-18
fix(Pico): ensure Pico loads on first page render

joemcgill:
This fixes a bug introduced in b590767 that caused Pico not to fire the initial page visit for the
initial pageview.

See: IRV-710

### DIFF
--- a/packages/integrations/components/pico/index.js
+++ b/packages/integrations/components/pico/index.js
@@ -15,7 +15,10 @@ import {
 } from '../../actions/coralActions';
 import { tokenSelector } from '../../selectors/coralSelector';
 // Utility functions.
-import mountPicoNodes from './utils';
+import {
+  isPicoMounted,
+  mountPicoNodes,
+} from './utils';
 
 const Pico = (props) => {
   const {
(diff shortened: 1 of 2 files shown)


packages/integrations/components/pico/index.js
```diff
@@ -83,21 +86,16 @@ const Pico = (props) => {
     const widgetContainer = document.getElementById('pico-widget-container');
 
     if (widgetContainer) {
-      // Prevent the widget script from being added more than once.
-      setPicoInitialized(true);
```
joemcgill, line 87: This is being handled as part of an event listener.

packages/integrations/components/pico/index.js
```diff
@@ -83,21 +86,16 @@ const Pico = (props) => {
...
       // Ensure the target nodes are only mounted once on the initial server load.
-      if (
-        ! document.getElementById('PicoSignal-container') &&
-        ! document.getElementById('PicoRule-button') &&
-        ! picoLoaded
-      ) {
+      if (! isPicoMounted()) {
         mountPicoNodes();
+        // Dispatch initial Pico page visit.
+        dispatchUpdatePicoPageInfo(picoPageInfo);
```
joemcgill, line 93: This should dispatch the initial Page action, which isn't getting fired on initial render.

ostowe (approved):
💯

PR #345, opened 2020-09-18
fix(pico): Reinsert initialization process

joemcgill:
This reverts part of the change in 86130c3.


joemcgill:
Self approving this since it's just reverting some changes that were already in place so I can test it on the alpha branch.

PR #346, opened 2020-09-21
fix(pico): Fix timing bug with firing the initial Pico action

joemcgill:
This changes the order of the initial dispatched actions after the Pico notes are mounted in ordre
to try and fix a race condition where the first payload to Pico is being fired too early.

IRV-710


tylerreckart (approved):
👍  This looks good on my local environment. I'd like to get an alpha release out to test on develop to ensure that this fixes the issue.

PR #348, opened 2020-09-22
IRV-710: Fix Pico load states

tylerreckart:
This PR fixes an issue where the Pico `pageInfo` would not be updated when navigating through article pages and viewing the same article more than once. This is done by removing an extraneous call to `dispatchPicoPageInfo` that was causing the action to be fired twice on an initial server render and adding the current URL as a dependency to the mount effect hook.


jameswburke (approved):
:bee:

PR #350, opened 2020-09-23
Refactor documentation

jameswburke:
## Note to reviewer
This PR does a lot. It's basically an entire overhaul and reorganization of the Storybook, including bringing in everything in both Irving and WP Irving's wikis.

I've pushed this entire thing to http://storybook-qa.irvingjs.com/ to help make UAT easier.

Docs are constantly improving, and there are a lot of `coming soon` or `this is out of date` messages, and that's okay.

IRV-590

* Refactor the documentation structure.
* Fill in some blanks from the repo wikis.
* Remove/update some outdated info.

Just docs.


stories/04-features/Authentication.stories.mdx
```diff
@@ -0,0 +1,18 @@
+<Meta title="Features|Authentication" />
+
+# Authentication
+
+## Basic Authentication
+Basic auth can be enabled by enabling two environment variables. They will automatically apply the Authorization header for requests to the app. This will prompt the user to enter the valid credentials to access the site.
```
ostowe, line 6: This actually requires one more env var now, `BASIC_AUTH=true`. This is because basic auth is also used to protect cache endpoints, so there has to be a way to do the two separately (protect endpoints vs. put the site behind basic auth).

stories/07-styling/02-css-modules-and-postcss.stories.mdx
```diff
@@ -0,0 +1,152 @@
+<Meta title="Styling|CSS Modules and PostCSS" />
```
ostowe, line 1: These shouldn't be in the same section, as a user could use CSS Modules without PostCSS. I would also like to create a SASS package at some point as well, which could be used with CSS modules.

ostowe (approved):
Left you a couple comments on a first pass, will probably do a second pass as well. This looks gooooooood thank you for putting in so much time on this reorg.

stories/07-styling/02-css-modules-and-postcss.stories.mdx
jameswburke: Discussed in Slack, going to leave as-is since you can only use PostCSS in the context of CSS modules.

PR #354, opened 2020-09-24
DEF-274 a11y remediation pagination

lmfern:
DEF-274

add aria-labels to the Pagination Wrapper and NextAndPrevNavWrapper so that the links have discernible names.

I tested it using the SiteImprove extension.


ostowe (approved):
💯  lgtm! 

For context (this is also for @jameswburke): pretty sure the package locks updated because you ran `npm run develop:prepare`, which will install dependencies for all packages, which in turn will update `package-lock.json` to use the latest patch version of each package.

If the package maintainers are using semver correctly, this should be completely safe (and is actually probably a good thing for us to do periodically).

PR #361, opened 2020-09-28
Abstract node polling functionality and modify it slightly

ostowe:
N/A

* Create `usePollForNode` hook.
* Modify polling functionality to use a different state determining if the node has been found—`null` if not found, the DOM node itself if found.

Needs to be tested on develop


joemcgill (approved):
👍

PR #363, opened 2020-09-29
fix(gtm): fix XSS issue in GTM

joemcgill:
This ensures the dataLayer is serialized inside the inline script to avoid URL based XSS attacks.

IRV-768


jameswburke (approved):
: 🐝

PR #365, opened 2020-09-30
IRV-745: Resolve issues with Pico load states

tylerreckart:


packages/integrations/components/pico/useWidgetScript.js
```diff
@@ -0,0 +1,37 @@
+/* eslint-disable max-len */
+import { useEffect, useCallback } from 'react';
+import { useDispatch, useSelector } from 'react-redux';
+import getLogService from '@irvingjs/services/logService';
+import { actionPicoScriptAdded } from '../../actions/picoActions';
+import { picoScriptAddedSelector } from '../../selectors/picoSelector';
+
+const log = getLogService('irving:pico');
+
+export default function useWidgetScript(widgetUrl, publisherId) {
```
ostowe, line 10: Seems like this could be a cool thing to abstract at some point too (functionality for adding an inline script, unlike `useLoadScript` which adds a script tag with `src`)

packages/integrations/components/pico/index.js
```diff
@@ -21,6 +25,8 @@ import {
   mountPicoNodes,
 } from './utils';
 
+const log = getLogService('irving:pico');
```
ostowe, line 28: I would actually maybe rename this namespace to `irving:integrations:pico`

ostowe (approved):
💯 Not sure I fully understand the reasons behind all this code, but it all looks good to me!

PR #370, opened 2020-10-05
feat(byline): ensure all delimiter text is wrapped with spans with classes

amyevans:

N/A

* Wrap each delimiter in a span if it isn't already
* Add classes so that each piece can be targeted for styling

Tested locally


joemcgill (approved):
🎩

PR #371, opened 2020-10-06
IRV-774: Fix Coral login state when embed is in the document viewport

tylerreckart:
This PR fixes an issue where the Coral login state would not be updated if a user logged into Pico with the Coral embed visible within the document's viewport. The issue was caused by the fact that effect hook in the `withPico` HOC was not be re-run when the token was cleared from the Redux store. This PR adds the token value as an effect dependency to ensure that the effect is run any time that value changes.

This PR also includes a large refactor to the Pico observer component's architecture. This work was spurred by @ostowe's work on the `pico-experiments` branch to bring the signal status into the Redux store in order to make observer side-effects more dependable. This is accomplished by bringing the value of the signal into the store by adding a `signal` branch to the pico branch of the integrations state and setting effect dependency constants based on the values that exist in that object. Because the effect constants are now being defined by a selector, any time that selector's value changes the observer will be forced to re-render and subsequently re-calculate those effect constants' values. This leads to more reliable side-effects in the observer and removes the dependence on local state.

__Important:__ Any developer that reviews this work should also pull the branch locally to go through the Pico/Coral SSO flow to do a round of functional QA prior to approving for merge.


joemcgill (changes requested):
Going to look at this more tomorrow, but I'm seeing a few odd things that I think we need to double check.

1) On my test account, I'm not seeing tier properly set by Pico even though my account shows the correct tier, which is not allowing me to test this properly.

2) When I manually click the Coral "Join the Conversation" button, I'm seeing the Coral modal pop up. This may just be a setting misconfigured in the develop coral environment, because I'm seeing the same thing on the develop server, but I want to be sure we've got this sorted.

tylerreckart:
@joemcgill I was able to confirm that there was a setting in the Lede dev Coal instance that was allowing registration through the comment form. I've disabled that, so it should resolve the second issue you were seeing. I'm curious as to why Pico isn't applying the tier properly. We should set up a swarm to go through this later today.

joemcgill:
Went through things again this morning and I was able to resolve both issues I was experiencing yesterday but adjusting settings in the Coral dev and Pico staging accounts I was testing with. So far, things are looking good.

joemcgill (approved):
New props-type validation is working great. Everything else is testing out well. 🦑

PR #375, opened 2020-10-09
Fix issue with spreading props into link wrapper component

ostowe:
Spreading `{...props}` into the link wrapper caused the `theme` prop in the `LinkWrapper` styled component to be overridden, causing `siteTheme` to break. This PR will fix this issue.

Tested in client projects.


amyevans (approved):
✨

PR #382, opened 2020-10-15
fix(coral): Enforce component re-renders on selector variable changes

tylerreckart:
This PR re-introduces a state variable, `canComment`, into the withPico HOC wrapper for Coral. Additionally, it adds the Irving log service to track the component's lifecycle for easier debugging.

The issue this corrects was due to the fact that there were instances where even when component variables pulled from the Redux store would update, this would not trigger the wrapper to re-render. This caused an issue where even though users would have a valid Pico payment tier and would have received a JWT token to log the user into the embed via SSO, the unauthenticated embed instance would persist in the DOM leaving the user logged out.

This behavior is corrected by the aforementioned state variable and a `useEffect` hook that uses the variables derived from the Pico signal, the Coral JWT token, and the status of the Coral user's username as dependencies. Any time one of these values changes the effect will be re-run, potentially changing the value of `canComment` and enforcing a component re-render to return the authenticated coral instance.

### DIFF
--- a/packages/integrations/components/coral/withPico.js
+++ b/packages/integrations/components/coral/withPico.js
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import { useSelector, useDispatch } from 'react-redux';
+import getLogService from '@irvingjs/services/logService';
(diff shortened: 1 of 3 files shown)


joemcgill (approved):
Looks right to me 🤞🏻

PR #385, opened 2020-10-21
Fix bug causing order of icons to not be respected

amyevans:

N/A

Address a bug in the Social Sharing component that was causing platforms to display in the default platform order, instead of the passed platform order

Tested in client project


cahdeemer (approved):
🍁

PR #387, opened 2020-10-22
IRV-827: Integrate multisite config into proof of concept

tylerreckart:
## Summary

This PR expands the functionality for the multisite proof of concept by accepting a configuration provided in the root of the Irving client of the shape:
```
const config = [
  {
    domain: "site1.irving.alley.test",
    vars: {
      "API_ROOT_URL": "https://irving-root.alley.test/site1/wp-json/irving/v1",
      "ROOT_URL": "https://site1.irving.alley.test",
    },
  },
  {
    domain: "site2.irving.alley.test",
    vars: {
      "API_ROOT_URL": "https://irving-root.alley.test/site2/wp-json/irving/v1",
      "ROOT_URL": "https://site2.irving.alley.test",
    },
  },
];

module.exports = config;
```

Note: This will only work if the root WP instance is configured to serve multiple sites through the WP JSON API.


packages/core/utils/universalEnv.js
```diff
@@ -1,5 +1,25 @@
...
-);
+/* eslint-disable  no-underscore-dangle */
+const getEnv = () => {
+  const env = Object.keys(process.env).length ? process.env : window.__ENV__;
+
+  // If the `MULTISITE_CONTEXT` property has been added to window.__ENV__ during
+  // the webpack build process, replace instances of ROOT_URL and API_ROOT_URL
+  // with the corresponding values in the multisite configuration.
+  if (env.MULTISITE_CONTEXT) {
+    const { host } = window.location;
```
ostowe, line 9: Will this work server-side because of `shimDom`?

packages/core/server/index.js
```diff
@@ -43,6 +43,27 @@ const irvingServerMiddleware = getValueFromFiles(
...
 irvingServerMiddleware.forEach((middleware) => middleware(app));
 
+const { multisiteConfig } = require('../config/paths');
+
+const multisiteContext = getValueFromFiles(
+  multisiteConfig,
+  [() => {}]
+);
+
+app.use((req, res, next) => {
```
ostowe, line 53: Because of the modifications you've made to `universalEnv` I'm wondering if this middleware is even necessary anymore, at least for our MVP. If anything, it actually seems dangerous to overwrite `process.env.API_ROOT_URL` on (potentially) every user request.

ostowe (commented):
A few initial thoughts/questions for you, looking good so far though!

packages/core/server/index.js
```diff
@@ -43,6 +43,27 @@ const irvingServerMiddleware = getValueFromFiles(
 );
 irvingServerMiddleware.forEach((middleware) => middleware(app));
 
+const { multisiteConfig } = require('../config/paths');
+
+const multisiteContext = getValueFromFiles(
```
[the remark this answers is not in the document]
tylerreckart: No that's just a remnant of the original implementation. I'll be switching that to use `requireMultisiteConfig`. That's a great idea for providing that to the environment to both!

packages/core/utils/universalEnv.js
tylerreckart: I believe so. I have confirmed that the page source on a server-side render does include the correct pre-loaded state.

packages/core/server/index.js
tylerreckart: Unfortunately we need both. The process property replacement in `universalEnv` is required to pass the correct `API_ROOT_URL` through to `fetchComponents`. Without having the replacement there, I was seeing an issue with the request to the components endpoint coming through as `https://site2.irving.alley.test/undefined/components?page=/&context=page`. The server-side replacement ensures that the pre-loaded state will be set correctly when page initially renders.

I've added an additional equality check for the current value of `process.env.ROOT_URL` against the current host's `ROOT_URL` variable and to only perform the replacement if they're different. That should prevent overlapping changes. There's still a lot about it that needs to be tweaked to get things performing just right, but at this point I think both instances may be required. Maybe it's worth breaking out the replacement into a utility function that could be called in both instances?

ostowe: I think you should be able to get around that by including the multisite config merging logic in a utility, like you said, and making sure it's called/used when the `irvingEnv` var is provided by webpack here: https://github.com/alleyinteractive/irving/blob/main/packages/core/config/webpack/plugins.js#L30. That's the var used to set `window.__ENV__`, which should prevent that `undefined` issue from happening.

tylerreckart: Unfortunately that didn't work for me. The issue with setting the multisite context in the `plugins` portion of the webpack build is that there is no context as to the current host, so the overwrite doesn't happen correctly in those instances. When `window.__ENV__` is instantiated in the template and becomes available in `universalEnv`, we have the host available through `window.location` to perform the replace client-side if the matched root is not already present.

tylerreckart: I ultimately had to re-add the root variable replacement in `universalEnv` to get this to work on client-side renders. I tried a number of things, but the biggest issue is that the host is unavailable to the server until the initial request is sent. This happens after the webpack modules are loaded, thus meaning that the server hostname cannot be passed into the webpack build system.

danbowles (approved):

ostowe (approved):
💯 looking good! figure we can revisit to try and consolidate those calls to `maybeReplaceRootVars` as a later iteration.

tylerreckart:
@ostowe Yep! That's definitely already on the radar. We have a ticket created to track that effort. It's not in the current sprint, but should be in the upcoming development cycle.

PR #388, opened 2020-10-26
fix(html): add button to list of allowed tags

amyevans:

N/A

Allow `button` tags to be rendered by components utilizing the `richText` config for sanitization.

Tested on a client project


cahdeemer (approved):
🍁

PR #389, opened 2020-10-26
LEDE-329: Update versions to new beta

davisshaver:


ostowe:
@davisshaver this shouldn't be necessary, lerna will handle version bumps like this on its own.

davisshaver:
Thanks Owen, I realized this but only after creating

PR #393, opened 2020-11-03
HOTFIX: Remove Coral embeds to prevent side effect buildup

davisshaver:
Noticed because of multiple GTM events firing as pageviews accumulated. Uses `useHook` cleanup to remove embeds.


goodguyry (approved):
🎸  Nice

PR #397, opened 2020-11-12
LEDE-535: exit click handler early to enable command+click tab opening

tylerreckart:

### DIFF
--- a/packages/styled-components/components/link/index.js
+++ b/packages/styled-components/components/link/index.js
@@ -49,14 +49,27 @@ const Link = (props) => {
   } = theme;
   const standardProps = useStandardProps(props);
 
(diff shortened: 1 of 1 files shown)


packages/styled-components/components/link/index.js
```diff
@@ -49,14 +49,27 @@ const Link = (props) => {
...
     });
+
+    // Allow the user to open the link in a new tab
+    if (
+      event.ctrlKey ||
+      event.shiftKey ||
+      event.metaKey ||
+      (event.button && 1 === event.button)
+    ) {
+      return;
```
danbowles, line 67: Could avoid disabling eslint rule by just returning `() => {}`.

danbowles (approved):

PR #399, opened 2020-11-13
fix(html.js): add link tags to allow list

amyevans:

N/A

Ensure link tags aren't stripped

Tested locally


isGabe (approved):
🐴

PR #400, opened 2020-11-20
docs(multisite.config.js): update docs to add page on multisite config

xmattus:

### DIFF
--- a/stories/02-configuration/05-Multisite-Support.stories.mdx
+++ b/stories/02-configuration/05-Multisite-Support.stories.mdx
@@ -8,7 +8,7 @@ A single instance of Irving is capable (since version 5.2.0) of serving multiple
 
 The `ROOT_URL` and `API_ROOT_URL` environment variables defined in the `.env` file are the default values for the frontend and backend API root URLs. Optionally, these can be overridden by creating a config file named `multisite.config.js` in the Irving project root. If present, it is possible to override the `ROOT_URL` and `API_ROOT_URL` values at run time for matching hostnames.
 
-When Irving is transpiled, only the `.env` file variables are available in the `process.env` object. Multisite support uses a wrapper object around the base environment variables. The overridden variables are only available through this method, which also requirs the `window.location.href` global to be set.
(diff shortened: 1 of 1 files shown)


stories/02-configuration/05-Multisite-Support.stories.mdx
```diff
@@ -0,0 +1,39 @@
+<Meta title="Configuration|Multisite Support" />
+
+# Multisite Support
+This section outlines how to configure Irving to support multiple different sites using a single instance.
+
+## Overview
+An single instance of Irving is capable (since version 5.2.0) of serving multiple different URLs. Irving will query a different backend API root URL conditionally based on the incoming hostname in the HTTP request. Each backend host should be configured to be compatible with the same component map.
```
amyevans, line 7:
```suggestion
A single instance of Irving is capable (since version 5.2.0) of serving multiple different URLs. Irving will query a different backend API root URL conditionally based on the incoming hostname in the HTTP request. Each backend host should be configured to be compatible with the same component map.
```

stories/02-configuration/05-Multisite-Support.stories.mdx
```diff
@@ -0,0 +1,39 @@
...
+
+The `ROOT_URL` and `API_ROOT_URL` environment variables defined in the `.env` file are the default values for the frontend and backend API root URLs. Optionally, these can be overridden by creating a config file named `multisite.config.js` in the Irving project root. If present, it is possible to override the `ROOT_URL` and `API_ROOT_URL` values at run time for matching hostnames.
+
+When Irving is transpiled, only the `.env` file variables are available in the `process.env` object. Multisite support uses a wrapper object around the base environment variables. The overridden variables are only available through this method, which also requirs the `window.location.href` global to be set.
```
amyevans, line 11:
```suggestion
When Irving is transpiled, only the `.env` file variables are available in the `process.env` object. Multisite support uses a wrapper object around the base environment variables. The overridden variables are only available through this method, which also requires the `window.location.href` global to be set.
```

amyevans (approved):
✨ Looks good to me!

PR #430, opened 2020-12-16
Ignore irvingjs.com dist files

joemcgill:
- Removes all the built files from inside the `dist` directory.
- Updates the `.gitignore` file to ensure no more built files get tracked.


isGabe (approved):
🐴

PR #432, opened 2020-12-17
fix(loadable): use prefetch instead of preload for async chunks

ostowe:
[**LEDE-528**](https://alleyinteractive.atlassian.net/browse/LEDE-528)

* Switch all async chunks to using a `rel="prefetch"` link instead of `rel="preload"`. Still not 100% sure why this is necessary, but it seems to resolve the issue.

* Tested on client project

### DIFF
--- a/packages/loadable/server/getAppTemplateVars.js
+++ b/packages/loadable/server/getAppTemplateVars.js
@@ -23,8 +23,8 @@ export default function getAppTemplateVars(templateVars) {
            *  loadable components seems to enforce
            */
           extractor.getLinkTags().replace(
-            /rel="preload"/gi,
-            'rel="prefetch"'
+            /(data-chunk="[^"]+"\srel=)("preload")(\sas="[^"]+")/gi,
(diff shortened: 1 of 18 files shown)


packages/loadable/server/getAppTemplateVars.js
```diff
@@ -16,8 +16,17 @@ export default function getAppTemplateVars(templateVars) {
...
+        script: () => extractor.getScriptTags(),
+        link: () => (
+          /**
+           *  @todo find a better way to get rid of the auto-preloading that
+           *  loadable components seems to enforce
+           */
+          extractor.getLinkTags().replace(
+            /rel="preload"/gi,
+            'rel="prefetch"'
+          )
```
jameswburke, line 28: Will this affect other scripts too? i'm concerned about it affecting something that comes through the Irving endpoint.

jameswburke (approved):
:bee:

jameswburke (approved):
:bee:

PR #433, opened 2020-12-18
perf(core): Upgrade sanitize-html to 2.3.0

joemcgill:
The latest release of sanitize-html is a much smaller package size, and should improve performance
overall.

We were previously using versions ^1.23.0, which was 1.62 MB unpacked. Version 2.3.0 is reported as being 67.2 kB since they removed built files (including polyfills) in version 2.0.


isGabe (approved):
🎉

PR #437, opened 2020-12-22
feat(groupchildren utility): adds a groupChildren uility to core

jameswburke:
This utility returns an object where the components are mapped by the group prop/config value.

https://alleyinteractive.atlassian.net/browse/IRV-726

Add a utility which maps an array of children into an object using each child's `group` prop.

## Example
```
const components = [
  {
    name: "irving/example-a",
    config: {
      "group": "first",
    },
  },
  {
    name: "irving/example-b",
    config: {
      "group": "first",
    },
  },
  {
    name: "irving/example-c",
    config: {
      "group": "second",
    },
  },
  {
    name: "irving/example-d",
  },
];

const groups = groupChildren(components);

console.log(groups);
```

Output:
```
{
  first: [
    {
      name: "irving/example-a",
      config: {
        "group": "first",
      },
    },
    {
      name: "irving/example-b",
      config: {
        "group": "first",
      },
    },
  ],
  second: [
    {
      name: "irving/example-c",
      config: {
        "group": "second",
      },
    },
  ],
  rest: [
    {
      name: "irving/example-d",
    }
  ]
}
```

This utility has been battle tested on a few projects, and has unit tests.

### DIFF
--- a/packages/core/utils/children.js
+++ b/packages/core/utils/children.js
@@ -49,3 +49,30 @@ export const filterChildren = (prop, value, children) => (
 export const filterChildrenByName = (componentName, children) => (
   filterChildren('componentName', componentName, children)
 );
+
+/**
+ * Map an array of children into an object using the `group` prop.`
+ *
+ * @param {array} children An array of React components.
+ * @returns {object} Mapping of children by group.
+ */
+export const groupChildren = (children) => {
+  const mapping = {
+    rest: [], // Array containing ungrouped children.
+  };
+
+  // If the `group` prop exists, and the group has been allow-listed, push the
+  // child component.
+  children.forEach((child) => {
+    const { group } = child.props;
+
+    if (group) {
+      const currentGroup = mapping[group] || [];
+      mapping[group] = currentGroup.concat(child);
+    } else {
+      mapping.rest = mapping.rest.concat(child);
+    }
+  });
+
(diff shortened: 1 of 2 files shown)


packages/core/utils/children.test.js
```diff
@@ -80,3 +81,74 @@ it('should filter an array of react elements by api component name', () => {
     },
   ]);
 });
+
+const testGroupChildrenElement = {
```
joemcgill, line 85: ❤️

packages/core/utils/children.js
```diff
@@ -49,3 +49,30 @@ export const filterChildren = (prop, value, children) => (
...
+
+/**
+ * Map an array of children into an object using the `group` prop.`
+ *
+ * @param {array} children An array of React components.
+ * @returns {object} Mapping of children by group.
+ */
+export const groupChildren = (children) => {
+  const mapping = {
+    rest: [], // Array containing ungrouped children.
```
joemcgill, line 61: I agree with @amyevans comment in channel that `ungrouped` would be a better name here, but overall this looks good and is really handy.

joemcgill (approved):
👍🏻

packages/core/utils/children.js
jameswburke: Updated 👍

PR #443, opened 2020-12-29
LEDE-678: Add data layer push for Pico signal update

davisshaver:

See https://alleyinteractive.atlassian.net/browse/LEDE-678

Adds a side effect to the Pico signal update action letting tag manager know that status has changed.


isGabe (approved):
🐴

PR #444, opened 2020-12-30
LEDE-678: Only pass payload, not whole object

davisshaver:

See https://alleyinteractive.atlassian.net/browse/LEDE-678

Updates Pico signal data layer message to just pass payload.


packages/integrations/sagas/picoSaga.js
```diff
@@ -108,6 +108,6 @@ function trackUpdatePicoSignal(signal) {
   window.dataLayer = window.dataLayer ?? [];
   window.dataLayer.push({
     event: 'irving.updatePicoSignal',
-    picoSignal: signal,
+    picoSignal: signal.payload,
```
kjbenk, line 111: Maybe add a check for this property before trying to access it to prevent potential JS error.

kjbenk (approved):
Address comment then ⛳

PR #447, opened 2021-01-14
Add bundle analyzer behind a cli flag, also add commander for core CLI

ostowe:
[**IRV-961**](https://alleyinteractive.atlassian.net/browse/IRV-961)

* Add commander to better organize and manage Irving's core CLI as it grows (and probably the CLI package, when we create it)
* Add an `--analyze` flag for the build command that will apply BundleAnalyzerPlugin

tested locally


renatonascalves (approved):
💯

PR #448, opened 2021-01-14
Add required functionality to support site-specific stylesheets

ostowe:
[**IRV-953**](https://alleyinteractive.atlassian.net/browse/IRV-953)

* Abstract a couple pieces of functionality and export them for use by projects (our instance of `MiniCSSExtractPlugin`, `createAssetTags`)
* Fix a couple eslint issues
* Pass site-specific env to template vars config function(s)

Tested in lede

### DIFF
--- a/packages/core/server/utils/getWebpackAssetTags.js
+++ b/packages/core/server/utils/getWebpackAssetTags.js
@@ -6,17 +6,47 @@ import getEnv from 'config/irving/getEnv';
 let runtimeSrc;
 
 /**
+ * Generate appropriate HTML tags for an array of chunk assets.
+ *
+ * @param {array} assets Array of webpack assets for a specific chunk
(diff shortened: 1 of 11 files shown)


packages/core/server/utils/getWebpackAssetTags.js
```diff
@@ -30,31 +60,17 @@ const getWebpackAssetTags = (clientStats, hostname) => {
   }
 
   // Include any other webpack assets that haven't been included yet.
-  Object.keys(chunks).forEach((chunkName) => {
+  tags = Object.keys(chunks).reduce((acc, chunkName) => {
```
pattiereaves, line 63: Could this be written as 
```js
 const tags = Object.keys(chinks).reduce((acc, chunkName) => { /* logic */ }, [])
```
?

pattiereaves (approved):
🦄

PR #451, opened 2021-01-22
refactor(pico): Update the Pico integration to support Pico Gadget

joemcgill:

LEDE-843

This refactors the Pico integration to simplify the handling of events and side effects when Pico loads so we are no longer waiting on specific DOM elements from Pico to be present before loading our integration logic. Updates:

* Load two event listeners, via `useEffect` hook, `pico-init` which is fired when the wrapper script is ready, and `pico.ready` which fires when Pico Gadget has fully loaded. These trigger callbacks that will update the state of `pico.isLoaded` and `pico.isReady`, respectively.
* Simplify logic for mounting our nodes for responding to Pico Signal logic so it's added to the DOM once and stays persistent.
* Simplify the logic for loading the Pico script
* Mount an effect that fires on the first page load and every time the URL updates that will dispatch an action that updates our pageInfo in Redux, and use a saga that responds to those changes to send payloads to Pico to track page visits.


ostowe (approved):
💯

PR #452, opened 2021-01-22
Fix runtime loading without loadable

ostowe:
N/A

* Webpack runtime was failing to load when loadable package wasn't installed. This PR will fix that issue.

tested locally, will test in develop


joemcgill (approved):
🧙🏻

PR #453, opened 2021-01-23
fix(cli): fix link-all command

joemcgill:

This fixes a bug introduced in 7e27c211 which removed the action for the link-all command.


ostowe (approved):
💯

PR #454, opened 2021-01-25
fix(pico): only update Pico page info after authorized requests finish

joemcgill:

LEDE-843

This checks the status of the `loading` state in Redux, using the `useLoading()` hook before firing
off `UPDATE_PICO_PAGE_INFO` actions to ensure components are rendered for Pico to respond to.
Otherwise a race condition will block Pico from loading when a visitor is logged into the back end.


ostowe (approved):
💯

PR #456, opened 2021-02-01
Allow sites to load site-specific assets via multisite config

ostowe:
[**IRV-979**](https://alleyinteractive.atlassian.net/browse/IRV-979)

Add config fields for `entry` and `head` to multisite config, allowing sites to add site-specific assets.

Added unit tests, tested locally.


packages/core/config/multisite/getEnv.js
```diff
@@ -0,0 +1,25 @@
...
+const getEnv = require('../../utils/universalEnv');
+
+/**
+ * A utility function that replaces the `ROOT_URL` and `API_ROOT_URL` values
+ * in the process.env object.
+ * @param {string} hostname - The hostname to search for.
+ * @returns {object} env object, with site-specific values if available.
+ */
+module.exports = (hostname = window.location.hostname) => {
+  const env = getEnv(); // eslint-disable-line no-underscore-dangle
```
kevinfodness, line 11: There doesn't appear to be an underscore dangle here. Can this comment be removed?
```suggestion
  const env = getEnv();
```

packages/core/config/multisite/getEnv.test.js
```diff
@@ -0,0 +1,22 @@
...
+
+describe('getEnv', () => {
+  it('merge site-specific keys into process.env and return the resulting object', () => {
+    const mergedEnv = getEnv('multisite-one.irving.test');
+    expect(mergedEnv).toMatchObject({
+      API_ROOT_URL: 'https://irving-multisite.test/api',
+      ROOT_URL: 'https://multisite-one.irving.test',
+      BASIC_AUTH_USERNAME: 'alley',
+      BASIC_AUTH_PASSWORD: 'interactive',
+      BASIC_AUTH: true,
```
kevinfodness, line 11: Since we actually use this as basic auth for some of our staging environments, it shouldn't be in source code in a public repo.

kevinfodness (approved):
Left some suggestions. See, in particular, the one about basic auth. 🍣

packages/core/config/multisite/getEnv.test.js
ostowe: @kevinfodness ah the username/pass? absolutely, I'll change those to something else.

PR #457, opened 2021-02-01
fix(integrations): fix Coral login error when JWT fails

joemcgill:

LEDE-945

This fixes an issue where Coral is failing to decode JWT payloads when the payload includes URL
encoded versions of base64 encoded strings by splitting the JWT token and string replacing URL
encoded characters of the payload with the base64 alternatives.

Have tested this flow locally against the Coral integration.

### DIFF
--- a/packages/integrations/components/coral/index.js
+++ b/packages/integrations/components/coral/index.js
@@ -43,6 +43,29 @@ const CoralEmbed = ({
   // https://reactjs.org/docs/hooks-rules.html
   const shouldPurgeUser = useSelector(purgeSelector);
 
+  /**
+   * Decode the payload of a JWT token so it becomes a base64 encoded string.
+   *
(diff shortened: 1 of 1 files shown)


packages/integrations/components/coral/index.js
```diff
@@ -43,6 +43,29 @@ const CoralEmbed = ({
...
+   *
+   * @param string token A base64url encoded JWT token.
+   * @returns string A JWT token with a base64 payload.
+   */
+  const base64UrlDecodeJWTPayload = (token) => {
+    // Split the token into parts so we can decode the payload.
+    const parts = token.split('.');
+
+    // Replace URL encoded characters with the expected base64 alternatives.
+    parts[1] = parts[1].replace(/-/g, '+').replace(/_/g, '/');
```
ostowe, line 64: 🙀

ostowe (approved):
💯

PR #458, opened 2021-02-01
fix(integrations): fix Coral login error when JWT fails

joemcgill:
This backports #457 to the 5.4 branch.

This fixes an issue where Coral is failing to decode JWT payloads when the payload includes URL
encoded versions of base64 encoded strings by splitting the JWT token and string replacing URL
encoded characters of the payload with the base64 alternatives.

LEDE-945


ostowe (approved):
💯

PR #459, opened 2021-02-02
fix(html.js): add back iframe and time tag support

kaitlinbolling:

Add back the `iframe` and `time` tags to supported HTML.

Tested locally.


joemcgill (approved):
Looks good to me. Let's backport this to the release/5.4 branch as well and then do a 5.4.1-rc.1, since we've not released 5.4.1 yet. ✨

PR #462, opened 2021-02-09
 Make `irving/head` a provider

ostowe:
[**IRV-977**](https://alleyinteractive.atlassian.net/browse/IRV-977)

* modify the head component so it can be used as a provider.
* allow providers to receive both React tree children and apiChildren to do with what they will.

Wrote/modified some unit tests, tested locally


kaitlinbolling (approved):
👍  looks good!

PR #464, opened 2021-02-10
add `nowrap` prop to text component to prevent rendering outer div

ostowe:
## issue(s): Relates to or closes...
[**IRV-922**](https://alleyinteractive.atlassian.net/browse/LEDE-922)

* Add `nowrap` prop to text component
* Render `nowrap` html content via `react-html-parser`

Tested on client project locally.


amyevans (approved):
✨  Perf

PR #468, opened 2021-02-15
Fix hot loading with mini-css-extract-plugin

ostowe:
N/A

* Support sass extension as an exclude for file-loader
* Add dev runtime chunk

Tested locally on client project.


jameswburke (approved):
:bee:

PR #471, opened 2021-02-19
Bugfix/route provider updates

ostowe:
N/A

* Allow providers to re-render with a new route key even on cached pages (for which `RECEIVE_COMPONENTS` action is never dispatched, only `FINISH_LOADING`)
* Revert head component to the way it was before https://github.com/alleyinteractive/irving/pull/462 (head provider isn't working as expected)

Tested locally on client project, will also test in develop.


jameswburke (approved):
:bee:

PR #472, opened 2021-02-24
Fix body wrapper and loader components so they're actually configurable

ostowe:
[**LEDE-1187**](https://alleyinteractive.atlassian.net/browse/LEDE-1187)

* Add the loader component to the component map.
* Deprecate/remove the `fullScreen` and `fullScreenBgColor` options for the loader component as they are too prescriptive and difficult to override. We can add it back in later as a theme, if necessary.
* Allow a theme for the `BodyWrapper` component to cascade down and theme the `Loader` and `DefaultLoader` components as well.
* Add functionality to `withThemes` to allow passing in a theme directly to a component as a prop. This prop theme will take precedence over a `themeName`.

Tested locally on client projects

### DIFF
--- a/packages/core/config/webpack/plugins.js
+++ b/packages/core/config/webpack/plugins.js
@@ -104,7 +104,6 @@ module.exports = function getPlugins(context, argv) {
           filename: '[name].css',
           chunkFilename: '[id].css',
         }),
-
       ];
 
     default:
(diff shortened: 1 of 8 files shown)


jameswburke (commented):
:bee:

jameswburke (approved):
:bee:

PR #478, opened 2021-03-23
Add visited flag in redux to prevent multiple visit calls on a single navigation

ostowe:
N/A

* Add a flag in pico redux state to say if a call to `pico('visit')` has been made since the last navigation. This should prevent duplicate visits.
* Remove lingering outdated state values.

I've tested locally.


jameswburke (approved):
:bee:

PR #491, opened 2021-04-15
Update signal in redux on load even if mutation observer doesn't see changes

ostowe:
N/A

We don't know which will come first: pico updating the signal div or our mutation observer attaching to it. This will ensure that if Pico updates the div before we attach the observer, the signal attributes still make it into redux.

Tested locally


jameswburke (approved):
:bee:

PR #492, opened 2021-04-16
Fix issue with blockEditor build assets being removed

ostowe:
N/A

* Prevents clean webpack plugin from removing built assets from the block editor build

tested locally


kaitlinbolling (approved):
🧹

PR #493, opened 2021-04-20
remove sanitize-html and replace with dompurify

ostowe:
[**LEDE-1456**](https://alleyinteractive.atlassian.net/browse/LEDE-1456)

* Remove `sanitize-html` which has caused us continuous issues, replace with `dompurify` which is smaller and more in line with the current Alley standard.
* Fix storybook (I realize this has nothing to do with the intent of this PR, but some package updates necessitated it for it to continue working)
* Replace our DOM shim with `jsdom` as we need it anyway for `dompurify`
* Add a `sanitizeHTML` field to `irving.config.js` to allow users to add/modify sanitization configuration

Tested locally and on lede


packages/core/utils/sanitizeHTML.js
```diff
@@ -0,0 +1,39 @@
...
+  {
+    ADD_TAGS: [
+      'iframe', 'link',
+    ],
+    ADD_ATTR: [
+      'rel', 'href', 'type', 'target', 'nohref', 'xml:lang', 'currenttime',
+      'duration', 'char', 'charoff', 'hspace', 'longdesc', 'vspace', 'autofocus',
+      'dirname', 'form', 'formaction', 'formenctype', 'formmethod',
+      'formnovalidate', 'formtaget', 'rules', 'abbr', 'axis',
+    ],
```
kaitlinbolling, line 34: I think the following are included/not needed here: `rel`, `href`, `type`

A couple I think need to be added: `aria-*`, `data-*`, `buffered`, `intrinsicsize`

Also, might be nice to alphabetize this list for readability.

packages/core/utils/sanitizeHTML.js
```diff
@@ -0,0 +1,39 @@
...
+  sanitizeHTML = require('dompurify');
+}
+
+module.exports = sanitizeHTML;
+module.exports.config = getValueFromConfig(
+  'sanitizeHTML',
+  {
+    ADD_TAGS: [
+      'iframe', 'link',
+    ],
```
kaitlinbolling, line 28: Maybe add `title` (we had this for some reason)

kaitlinbolling (approved):
👍  Looks good! Just left some comments for consideration (I was cross-checking our tags/attrs w/ https://github.com/cure53/DOMPurify/tree/e1c19cf6407d782b666cb1d02a6af191f9cbc09e/src, but I might not be looking in the right place!)

packages/core/utils/sanitizeHTML.js
ostowe: Looks like we had `rel`, `href`, and `type` in here for `link` tags, so I'm gonna keep those in for now. afaik ARIA and data attributes are supported out of the box, and configured separately. Will add the other two, though.

PR #494, opened 2021-04-23
return early from components saga if we aren't redirecting

ostowe:
N/A

Should fix the following server-side error seen in newrelic: `Cannot set property location of#<Window> which has only a getter`

Tested locally, will test on develop/preprod


jameswburke (approved):
:bee:

PR #497, opened 2021-04-28
Incorporate an optional Sentry transport into VIP Go logging service

ostowe:
[**LEDE-1460**](https://alleyinteractive.atlassian.net/browse/LEDE-1460)

* Fix issue with incorrect log levels
* Create a more abstract method of getting the correct monitor or log service inside and outside the build
* Pass an actual `Error` instance to monitor service `logError` method when an error is logged
* Incorporate an optional Sentry transport into VIP Go logging

Tested locally


jameswburke (approved):
:bee:

PR #523, opened 2021-05-18
Remove `react-transition-group` from loader component, core

ostowe:
Also update react-fast-refresh to beta version to fix issue with linked development

https://alleyinteractive.atlassian.net/browse/LEDE-1496

* Remove `react-transition-group` package from core, remove integration with `Loader` component.
* Update package locks to make webpack versions match up between packages
* Update `@pmmmwh/react-refresh-webpack-plugin` and `react-refresh` to fix issue with developing while irving packages are linked.

Tested locally


jameswburke (approved):
:bee:

PR #524, opened 2021-05-20
attempt to address issue with fetch called using URL without protocol

ostowe:
https://alleyinteractive.atlassian.net/browse/LEDE-1516, maybe https://alleyinteractive.atlassian.net/browse/LEDE-1519 also

* Use the `url-parse` package to check for a protocol instead of the current less-reliable regex test

Tested locally with the URL seemingly causing the issue in sentry.


amyevans (approved):
✨

PR #525, opened 2021-05-21
WIP:Hotfix/lede 1521/include url

justinwhall:

LEDE-1521

Allow passing urls to the sentry logging services as an object like: `{errorUrl: 'https://joinlede.com/whatever/?query=string'}`

- Sentry dashboard
- Updated and passing `generateLagInfo.test.js`


packages/core/services/fetchComponents.js
```diff
@@ -44,7 +45,8 @@ export async function fetchComponents(
   const timeout = setTimeout(
     () => {
       log.error(
-        new Error('Components: Components Endpoint fetch was aborted for taking too long. Increase the `FETCH_TIMEOUT` environment variable.') // eslint-disable-line max-len
+        new Error('Components: Components Endpoint fetch was aborted for taking too long. Increase the `FETCH_TIMEOUT` environment variable.'), // eslint-disable-line max-len
+        { url: format('%s%s%s', hostname, path, search) }
```
justinwhall, line 49: Changed course here a bit. Upon further investigation, the protocol isn't easily available here and other places where the logger is used. Instead checking each message, we can pass an object with with a `url` key.

ostowe: Definitely nbd if we don't log protocol, especially in this first iteration. However, one possibility here would be to [pull out `ROOT_URL`](https://github.com/alleyinteractive/irving/blob/65258cbf3a6f8e9db068da2ca5fe48a3ea9050f0/packages/core/services/fetchComponents.js#L33) from the [multisite vars config](https://github.com/JoinLede/lede/blob/production/themes/lede-common/client/irving/multisite.config.js#L210) as well (which should include the protocol).

packages/vip-go/services/generateLogInfo.js
```diff
@@ -1,9 +1,21 @@
 const { format } = require('util');
 
 module.exports = function generateLogInfo(method, messages) {
+  const hasUrl = messages.filter(
+    (message) => 'object' === typeof message && message.url
```
ostowe, line 5: could this be `messages.find` instead (resulting in a single value vs. an array)?

ostowe (approved):
A couple minor comments, but looks good to me!

ostowe (approved):
💯

PR #528, opened 2021-05-25
Update lede common irving app to adhere to (nearly) unmodified airbnb standards

ostowe:
Yes, but ticket TBD

* Update eslint standards for all irving core packages to adhere (nearly) 100% to the unmodified airbnb config.

Tested locally with lede.


packages/wordpress/components/blockStyles/blocks/heading.js
```diff
@@ -1,7 +1,7 @@
 import { css } from 'styled-components';
 import { link } from './utils';
 
-/* eslint-disable indent */
+/* eslint-disable indent, import/prefer-default-export */
```
justinwhall, line 4: I wonder if it's worth adding an `.eslint` file in `blockStyles` if this is a common rule(s) we are disabling.

```
{
    "rules": {
        "import/prefer-default-export": "off"
    }
}
```

justinwhall (approved):
🚢

packages/wordpress/components/blockStyles/blocks/heading.js
ostowe: hmmmm true, that might be helpful. this is pretty common throughout the lede theme too...maybe we should just disable across the board within lede?

ostowe: Oh wait, I already disabled it in irving core but didn't remove these eslint disable lines.

PR #531, opened 2021-05-27
Add irving config to disable SSR

ostowe:
N/A

Allow a user to disable server side rendering of markup (and use `ReactDOM.render` instead of `hydrate`)

tested locally

### DIFF
--- a/packages/core/server/serverRenderer.jsx
+++ b/packages/core/server/serverRenderer.jsx
@@ -138,7 +138,7 @@ export default function serverRenderer(options) {
       log.error(
         '%o',
         { url: req.originalUrl, err },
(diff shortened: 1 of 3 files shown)


packages/core/server/serverRenderer.jsx
```diff
@@ -138,7 +138,7 @@ export default function serverRenderer(options) {
       log.error(
         '%o',
         { url: req.originalUrl, err },
-        { errorUrl: ROOT_URL + req.originalUrl, justin: 'test' },
```
justinwhall, line 141: 🤗

justinwhall (approved):

PR #533, opened 2021-05-27
Allow passing of arbitrary tags to sentry

ostowe:
https://alleyinteractive.atlassian.net/browse/LEDE-1592

* If you pass an object with a `tags` property to the log service, it'll use those as sentry tags.

tested locally


amyevans (approved):
✨

PR #543, opened 2021-06-02
Fix hash auto-scrolling for logged-in users

ostowe:
https://alleyinteractive.atlassian.net/browse/LEDE-1332

* Add a condition to the `waitToScroll` saga to check if user is logged-in
* Fix a bug with `useNavigationOnClick` hook to prevent error when component using the hook has children, and the user clicks on a child node.
* Also add tests to same hook.

Tested locally (on defector)


cahdeemer (approved):
🌸

PR #544, opened 2021-06-04
Feature/lede 1585/conditionally throw errors

justinwhall:

LEDE-1585

Adds an optional env var `IRVING_RENDER_ERRORS`. If set and `NODE_ENV` === `development`, Irving will not throw an _unhandled_ exception and crash but rather render the ejs error template and send the proper status code.


packages/core/server/serverRenderer.jsx
```diff
@@ -36,6 +36,9 @@ const render = async (req, res, clientStats) => {
   // Set up multisite env as early as possible
   const env = createClientEnv(req.hostname);
 
+  // Check to see if we are rending error to the browser.
```
amyevans, line 39:
```suggestion
  // Check to see if we are rendering errors to the browser.
```

amyevans (approved):
✨ looks good!

PR #546, opened 2021-06-07
Hotfix/irv 1012/multisite config 404

justinwhall:

IRV-1012

Prevents invalid fetch requests to the components endpoint in the event that `API_ROOT_URL` is undefined.

[Updated the fetchComponets.test.js.](https://github.com/alleyinteractive/irving/pull/546/files#diff-4feb3f0253d52995ee0e99434be85bea4499b32b531603a8b670c65efd1026a7R28)

### DIFF
--- a/packages/core/components/app/index.jsx
+++ b/packages/core/components/app/index.jsx
@@ -16,7 +16,7 @@ const App = (props) => {
   const { apiValid, error } = props;
 
   /**
-   * If the API URL is invalid, return early as only things bad happen from here.
+   * If the API URL is invalid, return early as only bad things happen from here.
    */
   if (!apiValid) {
     return (
--- a/packages/core/components/app/index.jsx
+++ b/packages/core/components/app/index.jsx
@@ -8,11 +8,23 @@ import getTrackingService from 'services/trackingService';
 import { getValueFromConfig } from 'config/irving/getValueFromConfig';
 
 const ErrorMessage = getComponent('error-message');
+const RouteInvalid = getComponent('route-invalid-message');
 const AppContentComponent = getComponent('app');
 const trackingService = getTrackingService();
 
 const App = (props) => {
-  const { error } = props;
+  const { apiValid, error } = props;
+
+  /**
+   * If the API URL is invalid, return early as only things bad happen from here.
+   */
+  if (!apiValid) {
+    return (
+      <ErrorBoundary>
+        <RouteInvalid />
+      </ErrorBoundary>
+    );
+  }
 
   return (
     <ErrorBoundary>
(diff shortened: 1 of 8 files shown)


packages/core/components/app/index.jsx
```diff
@@ -8,11 +8,23 @@ import getTrackingService from 'services/trackingService';
...
-  const { error } = props;
+  const { apiValid, error } = props;
+
+  /**
+   * If the API URL is invalid, return early as only things bad happen from here.
+   */
+  if (!apiValid) {
+    return (
+      <ErrorBoundary>
+        <RouteInvalid />
```
ostowe, line 24: Ooo, this is nice

packages/core/services/fetchComponents.js
```diff
@@ -23,9 +23,24 @@ export async function fetchComponents(routeMeta, routeCookies) {
...
+  const { FETCH_TIMEOUT, API_ROOT_URL } = env;
   const apiUrl = createEndpointUrl(routeMeta, routeCookies);
 
+  /**
+   * If for some reason the API_ROOT_URL is not defined or a user/bot finds
+   * an URL that is not mapped to a site on the network, return early
+   * and avoid making and invalid fetch request ex:(https://undefined/foo/bar).
+   */
+  if (!API_ROOT_URL) {
+    return {
```
ostowe, line 35: I think you could include a `log.warn` here if you wanted too. It would only cause noise locally, and even then probably only rarely (or someone could just filter out those logs)

ostowe (approved):
💯

packages/core/components/app/index.jsx
```diff
@@ -8,11 +8,23 @@ import getTrackingService from 'services/trackingService';
...
+const RouteInvalid = getComponent('route-invalid-message');
 const AppContentComponent = getComponent('app');
 const trackingService = getTrackingService();
 
 const App = (props) => {
-  const { error } = props;
+  const { apiValid, error } = props;
+
+  /**
+   * If the API URL is invalid, return early as only things bad happen from here.
```
justinwhall, line 19:
```suggestion
   * If the API URL is invalid, return early as only bad things happen from here.
```

packages/core/services/fetchComponents.js
```diff
@@ -23,9 +23,24 @@ export async function fetchComponents(routeMeta, routeCookies) {
...
   } = routeMeta;
   const env = getEnv(hostname);
-  const { FETCH_TIMEOUT } = env;
+  const { FETCH_TIMEOUT, API_ROOT_URL } = env;
   const apiUrl = createEndpointUrl(routeMeta, routeCookies);
 
+  /**
+   * If for some reason the API_ROOT_URL is not defined or a user/bot finds
+   * an URL that is not mapped to a site on the network, return early
+   * and avoid making and invalid fetch request ex:(https://undefined/foo/bar).
```
justinwhall, line 32:
```suggestion
   * and avoid making an invalid fetch request ex:(https://undefined/foo/bar).
```

PR #547, opened 2021-06-08
fix waitToScroll functionality to ensure it fires on server side loads

ostowe:
https://alleyinteractive.atlassian.net/browse/LEDE-1332 I guess

* Ensure hash is available, even on SSR
* set up `waitToScroll` saga on `FINISH_LOADING` and `RECEIVE_COMPONENTS` instead of `LOCATION_CHANGE` and set it up before auth request to ensure it's fired at the right times.

tested locally


justinwhall (approved):
🚢

PR #551, opened 2021-06-10
feature/LEDE-1596/pico-conversion

justinwhall:

### DIFF
--- a/packages/integrations/components/googleTagManager/index.jsx
+++ b/packages/integrations/components/googleTagManager/index.jsx
@@ -62,6 +62,12 @@ const GoogleTagManager = (props) => {
           {`
             window.dataLayer = window.dataLayer || [];
             if (${isNode()}) {
+              window.gtag = function () {
+                console.log(arguments);
+                window.dataLayer.push(arguments);
+              };
+              window.gtag('js', new Date());
+              window.gtag('config', ${containerId});
               var data = ${serialize(dataLayer, { json: true })};
               data.event = 'irving.initialRender';
               window.dataLayer.push(data);


packages/integrations/components/pico/usePicoGTMEvents.js
```diff
@@ -0,0 +1,44 @@
...
+];
+
+// Add listeners for Pico events.
+const usePicoGTMEvents = (gtmContainerId) => {
+  /**
+   * Return early if we don't have a containerId
+   * or gtag is not initialized.
+   */
+  if (!window || !window.gtag || !gtmContainerId) {
+    return null;
```
ostowe, line 19: Think there's ever a chance this hook will run, return null, then reach the `useEffect` on a subsequent run? Might cause React to yell at us if so

packages/integrations/components/pico/usePicoGTMEvents.js
```diff
@@ -0,0 +1,44 @@
...
+const log = getLogService('irving:integrations:pico');
+const events = [
+  'pico.user.registered',
+  'pico.user.authenticated',
+  'pico.payment.paid',
+  'pico.paymentMethod.added',
+];
+
+// Add listeners for Pico events.
+const usePicoGTMEvents = (gtmContainerId) => {
```
ostowe, line 13: Nice, this def makes more sense as a hook, good idea.

packages/integrations/components/pico/usePicoGTMEvents.js
ostowe: Thinking of this section: https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level

ostowe (approved):
One thing to consider, but looks great! 💯

packages/integrations/components/pico/usePicoGTMEvents.js
justinwhall: That is an excellent point!

PR #554, opened 2021-06-11
fix(pico): ensure Read More break selector data is passed to Pico correctly

amyevans:

N/A

Fix the Pico "Read More" WordPress block integration so that if an editor uses the `core/more` block to designate what content should be behind the paywall, that break is respected.

Tested locally in project


jameswburke (approved):
:bee:

PR #562, opened 2021-06-17
Remove call to next() which is causing response to end before write is finished

ostowe:
https://alleyinteractive.atlassian.net/browse/LEDE-1622

* Remove `return next()` from end of `purgeCache` middleware.

tested locally


justinwhall (approved):
🚢

PR #577, opened 2021-07-07
feat(a11ly): ➜ Pass aria atts and tabindex to link

justinwhall:

- Convert aria attributes to kebab case before spreading as camel case aria attributes are no supported.
- Add support for passing a tabindex prop.


cahdeemer (approved):
🌻

PR #582, opened 2021-07-13
Update deps and moves devDeps to devDeps

justinwhall:
Update deps and moves devDeps to devDeps


jameswburke (approved):
:bee:

Let's make sure we test this thoroughly on develop & preprod, we've been burned by dev dependencies before.

PR #587, opened 2021-07-22
feat(sentry): send response, request url and status to sentry 

justinwhall:

Send response, request URL and status to Sentry when `response.json()` fails.


amyevans (approved):
✨

PR #588, opened 2021-07-23
Fix bug with `util` package breaking in the browser

ostowe:
See https://github.com/browserify/node-util/issues/43

* Add fallback for `util` to irving core (in addition to the `vip-go` package, which already has it)
* Add `ProvidePlugin` to client builds to provide the `process/browser` package

Tested locally


justinwhall (approved):
🚢

PR #589, opened 2021-07-28
feat(commentcountstatic): add new Coral component for displaying static comment count data

amyevans:

N/A

* Add a new component that will render a static comment count passed to it (using markup that matches the output from the dynamic Coral comment count component)

Tested locally, and will test in preproduction environment.


jameswburke (approved):
:bee:

PR #590, opened 2021-07-28
fix(commentcountstatic): add support for additional Coral config in static component

amyevans:

N/A

Add `articleUrl` and `noText` config to achieve full parity with the original Coral comment count component. Most notably, the `data-coral-url` attribute needs to be present, otherwise Coral assumes the canonical URL and live-updates all Coral comment counts on the page to the same comment count when a new comment is made.

Tested locally and in preproduction environment


jameswburke (approved):
:bee:

PR #592, opened 2021-07-29
fix(commentcountstatic): differentiate class name for secondary Coral comment count component

amyevans:

N/A

Change the class name so that the 2 Coral comment count components are differentiated and the Coral embed does not override comment counts (`.coral-count` is a selector with significance for the embed)

tested locally


jameswburke (approved):
:bee:

PR #594, opened 2021-08-09
feature/LEDE-1787-isomorphic-logging

justinwhall:

- Refactors the logging service to support isomorphic logging.
- Adds @sentry/node
- Addes @sentry/react
- Removes `winston sentry transport`.
- Adds debug package (and uses for both server and client logging)

updated tests.

### DIFF
--- a/packages/core/config/irving/createClientEnv.js
+++ b/packages/core/config/irving/createClientEnv.js
@@ -15,14 +15,17 @@ const createClientEnv = (hostname) => {
   const allowlistArray = getValueFromConfig(
     'clientEnvAllowlist',
     [
-      'NODE_ENV',
       'API_ROOT_URL',
-      'DEBUG',
-      'ROOT_URL',
-      'ROUTE_COOKIES',
       'COOKIE_ALLOWLIST',
+      'DEBUG',
       'FETCH_TIMEOUT',
       'IRVING_EXECUTION_CONTEXT',
+      'IRVING_RENDER_ERRORS',
+      'NODE_ENV',
+      'ROOT_URL',
+      'ROUTE_COOKIES',
+      'SENTRY_DSN',
+      'SENTRY_ENVIRONMENT',
     ],
   );
   const allowlist = [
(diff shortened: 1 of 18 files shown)


packages/core/config/irving/createClientEnv.js
```diff
@@ -15,14 +15,17 @@ const createClientEnv = (hostname) => {
   const allowlistArray = getValueFromConfig(
     'clientEnvAllowlist',
     [
-      'NODE_ENV',
       'API_ROOT_URL',
-      'DEBUG',
-      'ROOT_URL',
-      'ROUTE_COOKIES',
       'COOKIE_ALLOWLIST',
+      'DEBUG',
```
justinwhall, line 20: Adds `IRVING_RENDER_ERRORS`, `SENTRY_DSN` & `SENTRY_ENVIRONMENT` and alpha sorts.

packages/vip-go/package.json
```diff
@@ -26,8 +26,7 @@
     "ioredis": "^4.16.3",
     "newrelic": "^5.13.1",
     "util": "~0.12.3",
-    "winston": "~3.3.3",
-    "winston-transport-sentry-node": "~1.0.2"
+    "winston": "~3.3.3"
```
ostowe, line 29: are we even using winston anymore after your changes here?

packages/vip-go/services/logService.js
```diff
@@ -15,100 +20,83 @@ const monitor = require('@irvingjs/core/services/monitorService/getService')();
...
-            }
+        // If we have an error, do a couple more things with it.
+        if (logInfo.level === 'error') {
+          const err = new Error(logInfo.message);
 
-            // Send error to production monitoring service.
-            if (NODE_ENV === 'production') {
-              monitor.logError(err);
-            }
+          if (logToSentry) {
```
ostowe, line 76: simpler than I thought it'd be, nice! 🥳

ostowe (approved):
💯  looks great, excited to get this up on prod and see the results! only future iteration I can think of here would be to move sentry into its own package/service or something so it's not tied to vip-go, but definitely something for down the road (if at all)

PR #607, opened 2021-08-19
Feature/lede 1825 sentry config

justinwhall:
LEDE-1825

1. Adds an alias for `logServiceConfig`
2. Adds support for adding a `sentry.config.js` (logServiceConfig) in the Irving `buildContext` directory. This allows an Irving app to pass an object to the Sentry `init` function. ex: https://github.com/JoinLede/lede/pull/1278/files
3. Refactors the `logService` to include the config object.

I went around and around on what to name things here as Sentry is baked into the log service. Suggestions on the front, if any, would be 👌.


jameswburke (approved):
:bee:

packages/core/config/irving/getServiceAliases.js
```diff
@@ -82,6 +82,17 @@ const getServiceAliases = (target) => {
     );
   }
 
+  /**
+   * Get the log service config.
+   */
+  let logServiceConfig = resolveConfigFilepath('sentry.config.js');
```
ostowe, line 88: Realized it might actually make more sense to put this in a `config/aliases.js` file in the `vip-go` package because it's only used by that package currently (assuming that's possible...I think it should be)

ostowe (approved):
had one comment to addess, but otherwise looks good!

packages/core/config/irving/getServiceAliases.js
justinwhall: I was looking at that and was going to bring it up in swarm actually! I'll give this a go.

ostowe (approved):
💯  🥳

PR #610, opened 2021-08-23
refactor(logging): move sentry config

justinwhall:

Moves the sentry to the route as `vip-go` package alongside `irving.config.js` as the `config` directory is removed prior to publishing to NPM which results in a broken build.


jameswburke (approved):
:bee:

PR #612, opened 2021-08-24
Refactor logging logic

justinwhall:

### DIFF
--- a/packages/vip-go/services/logService.js
+++ b/packages/vip-go/services/logService.js
@@ -8,6 +8,7 @@ const debug = require('debug');
 const generateLogInfo = require('./generateLogInfo');
 
 let initialized = false;
+let Sentry;
 
 /**
  * Create a debug logger that will conditionally handle logged errors based on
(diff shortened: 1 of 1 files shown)


packages/vip-go/services/logService.js
```diff
@@ -23,7 +22,6 @@ const getService = (namespace) => {
   let service = defaultService;
   let sentryConfig = {};
   let Sentry;
```
ostowe, line 24: I'd maybe suggest still putting this outside the function body, just in case.

ostowe (approved):
💯

packages/vip-go/services/logService.js
```diff
@@ -56,7 +54,7 @@ const getService = (namespace) => {
     sentryConfig = require('@irvingjs/config/sentryConfig');
   } else {
     const getSentryConfig = require('../config/getSentryConfig');
-    sentryConfig = getSentryConfig();
+    sentryConfig = require(getSentryConfig());
```
ostowe, line 57: Could maybe rename this function to `getSentryConfigPath` as well, for your future self.

packages/vip-go/services/logService.js
justinwhall: Oh right.

PR #613, opened 2021-09-01
fix(pico): Patch race condition by making sure pico script is loaded before triggering visit

justinwhall:

This PR makes sure the pico gadget script is loaded before calling `window.pico('visit', {...})`.

We were waiting on a number of conditions to dispatch a page visit (`dispatchUpdatePicoPageInfo`) one of which was `pico.scriptOnLoad`.  `pico.scriptOnLoad` is fired at the very beginning of the script and not after the script is loaded creating a brief race condition between the script loading and our action running.


jameswburke (approved):
:bee:

PR #616, opened 2021-09-07
fix(pico): fix local storage is not available

justinwhall:

Wait for pico instance before calling `getInstance()`.

### DIFF
--- a/packages/integrations/components/pico/index.jsx
+++ b/packages/integrations/components/pico/index.jsx
@@ -63,9 +63,10 @@ const Pico = (props) => {
 
   const irvingIsLoading = useLoading();
   const { ready: picoReady } = useSelector(picoLifecycleSelector);
-  const scriptLoaded = window?.Pico?.getInstance()?.scriptLoaded;
   const contentReady = useSelector(picoContentReadySelector);
   const visited = useSelector(picoVisitedSelector);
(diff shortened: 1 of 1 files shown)


packages/integrations/components/pico/index.jsx
```diff
@@ -63,9 +63,10 @@ const Pico = (props) => {
 
   const irvingIsLoading = useLoading();
   const { ready: picoReady } = useSelector(picoLifecycleSelector);
-  const scriptLoaded = window?.Pico?.getInstance()?.scriptLoaded;
   const contentReady = useSelector(picoContentReadySelector);
   const visited = useSelector(picoVisitedSelector);
+  const picoInstance = window?.Pico?.instance || false;
+  const scriptLoaded = picoInstance ? window?.Pico?.getInstance() : false;
```
amyevans, line 69: Does this still need to check for `.scriptLoaded`? Like

```suggestion
  const scriptLoaded = picoInstance ? picoInstance.scriptLoaded : false;
```

amyevans (approved):
✨

packages/integrations/components/pico/index.jsx
justinwhall: Yes. The instance in this case is sale, which was the root issue here.

PR #618, opened 2021-09-10
fix(coral): remove base64url encoding to base64 encoding conversion on Coral JWT token

amyevans:

N/A

Remove an earlier fix that converted the JWT token payload for Coral from base64url to base64. While this skirted the `atob()` error, it resulted in malformed tokens in some cases which would cause the Coral embed to die. The JWT spec calls for base64url, and Coral fixed the underlying issue in the Talk repo [here](https://github.com/coralproject/talk/pull/3206). Note that this requires you be on 6.3.7 or later of Coral for compatibility. 

Tested locally and in develop environments

### DIFF
--- a/packages/integrations/components/coral/index.jsx
+++ b/packages/integrations/components/coral/index.jsx
@@ -44,29 +44,6 @@ const CoralEmbed = ({
   // https://reactjs.org/docs/hooks-rules.html
   const shouldPurgeUser = useSelector(purgeSelector);
 
-  /**
-   * Decode the payload of a JWT token so it becomes a base64 encoded string.
-   *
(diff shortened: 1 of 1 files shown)


packages/integrations/components/coral/index.jsx
```diff
@@ -44,29 +44,6 @@ const CoralEmbed = ({
   // https://reactjs.org/docs/hooks-rules.html
   const shouldPurgeUser = useSelector(purgeSelector);
 
-  /**
```
ostowe, line 47: There's nothing more satisfying that cleaning up hacky code to handle edge-cases 🤩

ostowe (approved):
💯

PR #621, opened 2021-09-15
fix scriptLoaded flag to use pico lifecycle instead of accessing Pico object directly

ostowe:
https://alleyinteractive.atlassian.net/browse/LEDE-1924

* Rework code intended to prevent pico's `'Local storage not available'` error message to use the Pico lifecycle value instead of accessing the global `Pico` object directly.
* hoping this works better as it avoids calling `window.Pico.getInstance()` which, if called too early, seems to throw that local storage error.

* Tested locally (so far)


amyevans (commented):
✨

PR #630, opened 2021-09-29
feat(static comment count component): adds a new data attribute indic…

jameswburke:
…ating the number of comments

None.

Add the comment count value as a data attribute, so it can be targeted via css.

Tested locally.


ostowe (approved):
💯

PR #634, opened 2021-09-30
Add service aliases to storybook babel config

ostowe:
N/A

* Fix storybook builds to include service aliases (which, when missing, will break storybook for any component utilizing components in the `styled-components` package)

TEsted locally


justinwhall (approved):
🦛

